PR #83, opened 2026-04-01
feat(install): add Swival auto-registration

jedisct1:
If a Swival config is found, automatically register codedb there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@
 | Structural outlines (functions, structs, imports)      | Multi-project support                    |
 | Reverse dependency graph                               | Remote indexing over SSH                  |
 | Atomic line-range edits with version tracking          |                                          |
-| Auto-registration in Claude, Codex, Gemini, Cursor     |                                          |
+| Auto-registration in Claude, Codex, Gemini, Cursor, Swival |                                      |
 | Polling file watcher with filtered directory walker    |                                          |
 | Portable snapshot for instant MCP startup              |                                          |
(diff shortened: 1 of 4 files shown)


justrach (changes requested):
Thanks for the Swival integration work.

Before merge I still want the two install-script hardening fixes from the review thread:

- escape or otherwise safely serialize `codedb_bin` before writing it into the TOML command field
- create/update the Swival config with restrictive permissions instead of relying on the default umask-derived mode

I know similar raw writes exist elsewhere, but this PR is extending installer registration in a security-sensitive path, so I do not want to add another unescaped write or broad-permission config creation here. Once those two points are addressed, please tag `@codex review` again.

jedisct1:
@codex review

jedisct1:
Thank you @justrach !

I wanted to stay consistent with the existing code, but indeed, this is not a good direction.

Your suggestions have been applied.

Thanks!

justrach:
Closing because this PR predates the 0.16 migration ([#293](https://github.com/justrach/codedb/pull/293), merged 2026-04-18) which touched core layers (storage in #283, discovery in #284, MCP/server in #285). A lot has changed and the diff no longer applies cleanly.

Please rebase onto current main and reopen — happy to review the rebased version. Thanks for the contribution!

jedisct1:
Rebased, but no changes were needed, the diff doesn't touch any of the core layers.

jedisct1:
And only you can reopen a closed PR :)

PR #106, opened 2026-04-02
feat: Add csharp support

unliftedq:
As title, Added basic C# support to the explorer.

### DIFF
--- a/src/explore.zig
+++ b/src/explore.zig
@@ -74,6 +74,7 @@ pub const Language = enum(u8) {
     rust,
     go_lang,
     php,
+    csharp,
     markdown,
     json,
     yaml,
@@ -90,6 +91,7 @@ pub fn detectLanguage(path: []const u8) Language {
     if (std.mem.endsWith(u8, path, ".rs")) return .rust;
     if (std.mem.endsWith(u8, path, ".go")) return .go_lang;
     if (std.mem.endsWith(u8, path, ".php")) return .php;
+    if (std.mem.endsWith(u8, path, ".cs")) return .csharp;
     if (std.mem.endsWith(u8, path, ".md")) return .markdown;
     if (std.mem.endsWith(u8, path, ".json")) return .json;
     if (std.mem.endsWith(u8, path, ".yaml") or std.mem.endsWith(u8, path, ".yml")) return .yaml;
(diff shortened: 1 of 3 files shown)


unliftedq:
> Hey @unliftedq, thanks for adding C# support — the parser itself looks solid (using directives, namespaces, classes, interfaces, structs, enums, delegates, methods, constants with modifier stripping). Appreciate the contribution!
> 
> One issue though — the PR includes ~262 lines of whitespace reformatting on existing methods (`indexFileInner`, `getContent`, `cloneOutline`, `getTree`, `getImportedBy`, `getHotFiles`, `rebuildDepsFor`). Per our [CONTRIBUTING.md](https://github.com/justrach/codedb/blob/main/CONTRIBUTING.md):
> 
> > Do not mix unrelated file churn into bug-fix or perf PRs
> 
> This reformatting creates massive merge conflicts with every other open PR (we just merged PHP support in #87 and are about to merge HCL in #109), and makes it very hard to review the actual C# changes.
> 
> Could you resubmit with **only the C# parser additions**? Specifically just:
> 
> * `csharp` added to the `Language` enum
> * `.cs` detection in `detectLanguage`
> * The `parseCSharpLine` function and its helpers (`stripCSharpModifiers`, `lastIdent`)
> * The dispatch line in `indexFileInner`
> * The C# tests in `tests.zig`
> * `isCommentOrBlank` update for csharp (if applicable)
> 
> No reformatting of existing code — keep the original indentation as-is. That way it'll rebase cleanly and we can get it merged quickly.
> 
> Thanks again for the work!

It should be fixed now, previous unrelated change is unexpectedly introduced by the zig language server of my local zig extension.

justrach (changes requested):
Thanks for adding C# support and for cleaning up the unrelated whitespace churn.

I still see two blocking parser correctness issues on the current head in `parseCSharpLine()`:

- lines starting with `using ` are still treated as imports unconditionally, so local `using var ...` / `using (...)` statements will pollute `outline.imports` and dependency results
- the generic parenthesis branch is still broad enough to classify ordinary call expressions as `.method` symbols instead of declarations

Please tighten the declaration guard for methods and distinguish `using` directives from in-method `using` statements, then add or update tests for both cases. After that this should be much closer.

PR #115, opened 2026-04-02
fix: prevent trigram index OOM on large repos

### LINKED ISSUE #105: The performance is terrible on large repo
I try to run codedb on a repo with more than 40000+ files, it takes about 8 minutes for index.
Also, any search/find will trigger the trigram index OOM error: `could not persist trigram index: error.OutOfMemory`

justrach:
## Summary
Fixes #105 — trigram index OOM on repos with 40k+ files.

- **64KB file size cap for trigram indexing** — files > 64KB still get outline + word index but skip trigram construction, which is the main memory hog (each byte position generates a hashmap entry)
- **15k file count cap** — `initialScan` auto-skips trigram for files beyond 15k to prevent unbounded memory growth
- **OOM-resilient `rebuildTrigrams`** — catches `error.OutOfMemory` per-file and logs a warning instead of silently failing the entire rebuild

Search still works on large repos via word index and outline; trigram just provides faster substring matching for the files that fit in budget.

## Test plan
- [x] New test `issue-105: large files skip trigram indexing to prevent OOM`
- [x] All existing tests pass (`zig build test` exit 0)
- [ ] Manual: test on a 40k+ file repo to verify no OOM


justrach:
@codex review

PR #180, opened 2026-04-06
feat: composable search pipeline — codedb_query (#168)

### LINKED ISSUE #168: feat: agent-defined search — let agents compose custom search pipelines
## Problem

Agents currently pick from a fixed menu of 17 MCP tools. For complex queries ("find all auth-related files that import Redis and were modified this week"), they chain multiple tool calls — each round-trip costs latency and tokens.

## Proposed Solution

Add a `codedb_query` tool that accepts a composable search pipeline:

```json
{
  "name": "codedb_query",
  "arguments": {
    "pipeline": [
      {"op": "find", "query": "auth"},
      {"op": "filter", "has_symbol": "Redis"},
      {"op": "sort", "by": "modified", "order": "desc"},
      {"op": "limit", "n": 10}
    ]
  }
}
```

### Pipeline operators

| Op | Description | Example |
|----|-------------|---------|
| `find` | Fuzzy file search | `{"op": "find", "query": "auth"}` |
| `search` | Content search (text/regex/word) | `{"op": "search", "query": "handleAuth", "mode": "word"}` |
| `filter` | Filter by symbol, language, path glob | `{"op": "filter", "lang": "python"}` |
| `outline` | Get outlines for matching files | `{"op": "outline", "compact": true}` |
| `sort` | Sort by score/modified/size/path | `{"op": "sort", "by": "score"}` |
| `limit` | Truncate results | `{"op": "limit", "n": 5}` |
| `read` | Read matching files (with line ranges) | `{"op": "read", "lines": "1-50"}` |

### Why this matters

- **Token efficiency** — one round-trip instead of 3-5 chained tool calls
- **Agent autonomy** — agents compose the search they need instead of picking from a fixed menu
- **Consolidation** — could eventually replace several single-purpose tools
- **Inspired by** fff.nvim's approach of exposing just 3 flexible tools instead of many specialized ones

### Interaction with existing tools
- Existing tools remain as shortcuts (backwards compatible)
- `codedb_query` is the power-user tool for agents that want to compose
- `codedb_bundle` is similar but just batches independent calls — `codedb_query` chains them as a pipeline where each step feeds the next

### Files to modify
- `src/mcp.zig` — add `codedb_query` tool, pipeline parser, executor
- `src/explore.zig` — may need a pipeline execution context struct

justrach:
## Motivation

Agent performance depends not just on the model, but on the **harness** — the code that determines what information to store, retrieve, and present. Current MCP tool design forces agents into a multi-round-trip pattern: call `codedb_find`, read the results, call `codedb_search` on those files, read again, call `codedb_outline`. Each round-trip costs latency (~100ms) and tokens (the agent must parse, decide, and re-request).

This is the same problem identified in [Meta-Harness](https://yoonholee.com/meta-harness/) (Lee et al., 2026): compressed feedback across multiple tool calls loses the information needed to trace downstream failures to earlier decisions. A single `codedb_query` pipeline gives the agent **full access to intermediate results within one call** — the equivalent of giving the harness richer access to prior experience in fewer iterations.

Key insight: instead of the agent acting as the pipeline orchestrator (5 tool calls, each losing context), `codedb_query` moves the pipeline **inside the tool** where each step has zero-cost access to the previous step's full output.

## Design

`codedb_query` accepts a pipeline array where each step operates on the file set from the previous step:

```json
{"pipeline": [
  {"op": "find", "query": "auth"},
  {"op": "filter", "ext": ".py"},
  {"op": "limit", "n": 5},
  {"op": "outline"}
]}
```

### Pipeline ops

| Op | Input | Output | Description |
|----|-------|--------|-------------|
| `find` | — | file set | Fuzzy file search (Smith-Waterman) |
| `search` | optional file set | file set + matching lines | Content grep; intersects with prior set if present |
| `filter` | file set | filtered file set | Filter by `ext` (extension) or `glob` (path pattern) |
| `outline` | file set | symbol outlines | Get structural outline for each file in set |
| `read` | file set | file contents | Read file contents (with `lines` limit per file) |
| `sort` | file set | sorted file set | (Reserved for future: sort by score/path/size) |
| `limit` | file set | truncated file set | Keep only first N files |

### Interaction with existing tools
- Existing tools remain unchanged (backwards compatible)
- `codedb_bundle` batches independent calls in parallel — `codedb_query` chains them as a pipeline
- Agents can use `codedb_query` for complex multi-step queries and individual tools for simple ones

## Test plan
- [x] 7 new tests: find→limit, search with proper cleanup, filter by extension, outline with symbol verification, find→filter narrowing, empty results
- [x] All 309 tests pass (zero leaks)
- [x] Live MCP test: find→limit→outline, find→filter→limit, search→limit→outline
- [x] Pre-push benchmarks pass

Closes #168

### DIFF
--- a/src/mcp.zig
+++ b/src/mcp.zig
@@ -260,6 +260,7 @@ pub const Tool = enum {
     codedb_projects,
     codedb_index,
     codedb_find,
+    codedb_query,
 };
 
 const tools_list =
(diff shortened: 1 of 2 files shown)


justrach:
All three Codex review findings addressed:

**P1 — Search intersection**: Fixed. The `search` step now builds a `hit_set` of matched paths and compacts `file_set` to only those files. Downstream ops like `outline`/`read` now correctly operate on the narrowed set.

**P2 — Search allocation leak**: Fixed. Added proper `defer` block that frees `r.line_text` and `r.path` for each result, matching the pattern in `handleSearch`. Additionally fixed a **use-after-free** (caught by local Codex review): when search is the first pipeline step, paths are now `dupe`d before storing in `file_set`, since the defer frees the originals.

**P2 — Glob matching**: Fixed. Replaced `indexOf` substring check with proper `globMatch` function supporting `*` (single segment), `**` (cross-segment), and `?` (single char, not `/`). The `?` wildcard correctly refuses to match path separators.

All 319 tests pass, zero leaks.

PR #193, opened 2026-04-07
ci: add GitHub Actions workflow for Zig tests

unliftedq:
Introduced a new test pipeline to make sure there's no unexpected regressions.


justrach (changes requested):
Thanks for wiring up the Zig test workflow. This is useful.

One blocking issue before merge: both Zig install branches currently download archives from the network and immediately extract/run them without any integrity verification. Please pin and verify the archive checksum (or another equivalent integrity mechanism) for both the Unix and Windows downloads before extraction.

Once that is in, rerun the workflow and tag `@codex review` again.

PR #214, opened 2026-04-07
fix: release ProjectCache contents after snapshot load

### LINKED ISSUE #210: bug: ProjectCache retains raw file contents — 4.5GB leak on large repos
## Summary

`ProjectCache` in `src/mcp.zig` loads snapshots for non-default repos but never calls `releaseContents()` or `releaseSecondaryIndexes()` on the cached `Explorer`. This retains raw file contents in memory for the lifetime of the cache entry.

On a ~22K file repo, this causes **4.5 GB RSS** from a single cached project.

## Root Cause

In `mcp.zig` `ProjectCache.get()` (lines 90-127), after `loadSnapshot()` succeeds, the new `Entry` is inserted into the cache with full file contents still in the Explorer's `contents` HashMap.

Compare with the primary explorer in `main.zig` `scanBg()` (lines 794-797), which correctly releases contents after loading:

```zig
// main.zig — primary explorer does this:
if (fc > 1000 or std.process.hasEnvVarConstant("CODEDB_LOW_MEMORY")) {
    explorer.releaseContents();
    explorer.releaseSecondaryIndexes();
}
```

**`ProjectCache.get()` never calls either of these.** The raw file contents from the snapshot remain in memory for as long as the cache entry lives (up to `MAX_CACHED = 5` entries).

## Reproduction

```bash
# Start MCP server for a small default repo
codedb mcp /path/to/small-repo &
PID=$!

# Query a large repo (~22K files) via the project parameter
echo '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"codedb_status","arguments":{"project":"/path/to/large-repo"}}}' | ...

# Check RSS — should be ~100MB, actually ~4.5GB
```

## Measurements

| State | RSS |
|-------|-----|
| After startup (small default repo) | 89 MB |
| After first query to ~22K file repo | ~4,500 MB |
| Expected (with releaseContents) | ~200-300 MB |

## Fix

After `loadSnapshot` succeeds in `ProjectCache.get()`, release contents that aren't needed for query serving:

```zig
// After line 127 (successful load), before finding cache slot:
const fc = new_entry.explorer.outlines.count();
if (fc > 1000) {
    new_entry.explorer.releaseContents();
    new_entry.explorer.releaseSecondaryIndexes();
}
```

The snapshot contains outlines + trigram index + word index — sufficient for all MCP query tools (`search`, `symbol`, `outline`, `read`, `deps`, etc.). Raw file contents are only needed during initial indexing, not for query serving from a pre-built snapshot.

## Related

- #128 (closed) — primary explorer memory leak (fixed with mmap + releaseContents for the main repo). This is the **same class of bug** but in the multi-project cache path, which was not covered by the #128 fix.
- #207 — CLOCK eviction cache (would help limit total cached projects, but doesn't fix the per-entry content retention)

octo-patch:
Fixes #210

## Problem
`ProjectCache.get()` in `src/mcp.zig` loads snapshots for non-default repos but never calls `releaseContents()` or `releaseSecondaryIndexes()` on the cached `Explorer`. This retains raw file contents in memory for the lifetime of each cache entry. On a ~22K file repo, this causes ~4.5GB RSS from a single cached project.

## Solution
After `loadSnapshot` succeeds, release file contents that are not needed for query serving (the snapshot provides outlines + trigram index + word index, which is sufficient for all MCP query tools). Mirrors the identical pattern already used by the primary explorer in `main.zig` `scanBg()`.

## Testing
The fix mirrors the existing `main.zig` pattern exactly (`fc > 1000 or CODEDB_LOW_MEMORY`). Verified by code inspection that `releaseContents` and `releaseSecondaryIndexes` are safe to call after snapshot load — the snapshot contains outlines and indexes sufficient for all query tools.

### DIFF
--- a/src/mcp.zig
+++ b/src/mcp.zig
@@ -124,6 +124,12 @@ const ProjectCache = struct {
             }
         }
 
+        const fc = new_entry.explorer.outlines.count();
+        if (fc > 1000 or std.process.hasEnvVarConstant("CODEDB_LOW_MEMORY")) {
+            new_entry.explorer.releaseContents();
+            new_entry.explorer.releaseSecondaryIndexes();
+        }
+
         // Find free slot or evict LRU
         var target_slot: usize = 0;
         var found_free = false;


justrach:
Thanks for the PR. This is a useful fix.

There are still a few Codex review comments to address before merge.

Can you please go through that feedback and push the requested changes?

Once you have updated the PR, tag `@codex` so it can re-review the branch.

justrach (changes requested):
Thanks for the PR. This is a useful direction for the ProjectCache retention issue.

One blocking change before merge: `releaseContents()` cannot drop cached file contents before the cache has initialized the project root on the embedded explorer. On the current head, once contents are released, `codedb_read` / `codedb_search` fall back to disk reads via an explorer whose root is still unset, so non-default `project` roots can read relative to `cwd` and fail or silently miss matches.

Please make sure the project root is set on the cache entry before the contents-release path can run, then tag `@codex review` again once that is updated.

justrach:
Thanks for the contribution. The `#210` ProjectCache memory fix is already on `main`, so I am closing this PR as superseded rather than leaving two versions of the same change open. Appreciate you flagging and patching the leak.

PR #225, opened 2026-04-09
fix/issue 224 symbol line end

### LINKED ISSUE #224: bug: codedb_symbol body=true returns only the signature line — Symbol.line_end is never populated
## Problem

`codedb_symbol` with `body=true` is documented to return the full source of the matched symbol so agents can avoid a follow-up `codedb_read`. In practice it returns **only the signature line**.

Example against a real TypeScript file:

```
codedb_symbol { name: "handleResendWebhook", body: true }
```

Response:

```
1 results for 'handleResendWebhook':
  backend/src/services/resendWebhookService.ts:105 (function)
  // export async function handleResendWebhook(event, svixId): Promise<...> {
  105 | export async function handleResendWebhook(event, svixId): Promise<...> {
```

The actual function is 55 lines (L105–L159). Only line 105 is returned. Reproduced across `recordBounce`, `checkAndSuspend`, `handleBounced`, `validateNewsletterForSending`, `suspendBouncedSubscribers` — every multi-line function behaves the same way, regardless of language.

### Root cause

Every per-language parser in `src/explore.zig` writes:

```zig
.line_start = line_num,
.line_end   = line_num,
```

The `line_end` field exists on `Symbol` (`src/explore.zig:33`) since the first commit (`c7b28d4`, 2026-03-03), but it has **never** been populated with the real end of the block — it has always been set to the signature line. This affects `parseZigLine`, `parsePythonLine`, `parseTsLine`, `parseRustLine`, `parsePhpLine`, `parseGoLine`, and `parseRubyLine` (~35 occurrences of `.line_end = line_num`).

The `body=true` flag was added in `8051d12` (2026-03-05, "feat: add agent-friendly MCP enhancements"), wiring `codedb_symbol` through `explorer.getSymbolBody(path, line_start, line_end, ...)` in `src/mcp.zig:783`. That call assumes a correct `line_end` and extracts `[line_start..line_end]` via `extractLines` — which, with `line_end == line_start`, returns a one-line slice. The feature commit did not add a regression test that would catch this.

Every language added after that commit (`feat: add Go and Ruby language support (#151)`, `feat(explore): add PHP/Laravel language support (#87)`, `feat: add Rust symbol parsing to codedb_outline`) replicated the same `.line_end = line_num` pattern, propagating the bug.

## Failing Test

Branch (on fork): [`issue-224-failing-test`](https://github.com/ocordeiro/codedb/commits/issue-224-failing-test/) · commit [`9aeca60`](https://github.com/ocordeiro/codedb/commit/9aeca60).

Test added to `src/tests.zig`:

```zig
test "issue-224: codedb_symbol body=true returns only signature — line_end never populated" {
    defer arena.deinit();
    const alloc = arena.allocator();

    var explorer = Explorer.init(alloc);

    // Multi-line function: signature on line 1, body on lines 2..4, closing brace on line 5.
    try explorer.indexFile("t.zig",
        \\pub fn foo() u32 {
        \\    const a: u32 = 1;
        \\    const b: u32 = 2;
        \\    return a + b;
        \\}
    );

    const results = try explorer.findAllSymbols("foo", alloc);
    defer alloc.free(results);
    try testing.expect(results.len == 1);

    const sym = results[0].symbol;
    try testing.expectEqual(@as(u32, 1), sym.line_start);
    // With the bug, line_end == line_start (== 1). After the fix, it must reach
    // the closing brace on line 5.
    try testing.expectEqual(@as(u32, 5), sym.line_end);

    // Full-body extraction via getSymbolBody — the exact path codedb_symbol body=true
    // takes — must contain every body line, not just the signature.
    const body = (try explorer.getSymbolBody("t.zig", sym.line_start, sym.line_end, alloc)) orelse
        return error.TestUnexpectedResult;
    try testing.expect(std.mem.indexOf(u8, body, "pub fn foo()") != null);
    try testing.expect(std.mem.indexOf(u8, body, "const a: u32 = 1;") != null);
    try testing.expect(std.mem.indexOf(u8, body, "const b: u32 = 2;") != null);
    try testing.expect(std.mem.indexOf(u8, body, "return a + b;") != null);
}
```

Reproduce from upstream `main`:

```bash
# From a clone of the upstream repo
git remote add ocordeiro https://github.com/ocordeiro/codedb.git
git fetch ocordeiro issue-224-failing-test
git checkout ocordeiro/issue-224-failing-test
zig build test 2>&1 | grep "issue-224"
```

Fails with:

```
error: 'tests.test.issue-224: codedb_symbol body=true returns only signature — line_end never populated' failed: expected 5, found 1
src/tests.zig:5397:5: in test.issue-224: ... (test)
    try testing.expectEqual(@as(u32, 5), sym.line_end);
```

All other 323 tests pass.

## Expected

For every multi-line symbol kind, `Symbol.line_end` must point to the closing delimiter of the block (last `}` for brace languages, last body line for Python, `end` keyword for Ruby). `codedb_symbol body=true` and `explorer.getSymbolBody(...)` should then return the full source of the symbol in a single call, eliminating the follow-up `codedb_read` that the flag was added to avoid.

Single-line kinds (`.import`, `.variable`, `.constant`, `.comment_block`, `.type_alias`, `.macro_def`) should remain with `line_end == line_start`.

## Fix

Post-processing pass in `indexFileInner` (`src/explore.zig`), inserted between the end of the streaming parse loop and the `self.mu.lock()` call (while `content` is still in scope). The parsers themselves are not modified — a single helper, `computeSymbolEnds(content, &outline)`, walks the already-populated `outline.symbols` and fills in `line_end` per language:

- **Brace languages** (Zig, C, C++, TypeScript, JavaScript, Rust, Go, PHP): scan forward from `line_start`, tracking brace depth and skipping `"..."` strings, `//` line comments and `/* */` block comments, until the first opened `{` is closed. Bails out after 10 lines without an opener so forward declarations / abstract methods stay single-line.
- **Python**: indent-based. First contiguous run of lines whose indent is strictly greater than the signature's. Tolerates multi-line signatures, blank lines, and comments inside the body.
- **Ruby**: indent-based plus standalone `end` snapping at signature indent.

Naturally single-line kinds are short-circuited. Struct/enum/union defs already use dedicated `.*_def` kinds, so they flow through brace scanning correctly.

This design was chosen over refactoring each `parseXLine`:

1. The parsers are streaming per-line — giving them access to the rest of the file would require restructuring all 7.
2. A single fix location covers every current language and any future one.
3. Nothing about symbol detection — which has test coverage — is touched.

Side effects checked:

- `extractLines` (`src/explore.zig`) already supports real `line_end` values (1-based, inclusive) — unchanged.
- `getSymbolBody` and the `codedb_symbol` handler in `src/mcp.zig:783` — unchanged.
- `codedb_outline` rendering reads only `line_start` — unaffected.
- Snapshots persisted on disk still contain the old `line_end == line_start` values; users will need to re-index to pick up correct ranges after the fix. Worth a note in the changelog.

Fix branch (on fork): [`fix/issue-224-symbol-line-end`](https://github.com/ocordeiro/codedb/commits/fix/issue-224-symbol-line-end/) · commit [`8919fb8`](https://github.com/ocordeiro/codedb/commit/8919fb8).

`zig build test --summary all`: **333/333 passing** with 9 new tests covering the regression plus one per language (Zig fn, Zig struct_def, Python def, Python class, TS function, Rust fn, Go func, PHP function, Ruby def) and an abstract-signature edge case.

### End-to-end verification

Smoke-tested against the fix binary by calling `codedb_symbol name=computeSymbolEnds body=true` via MCP — returned all 36 lines of the function body (L1939–L1974 in `src/explore.zig`), not just the signature line.

ocordeiro:
Fixes #224.

> **Disclaimer**: I'm not a Zig programmer. This patch was produced with [Claude Code](https://www.anthropic.com/claude-code) (Opus 4.6) — I designed the approach, reviewed the code and test output, and validated the fix end-to-end via MCP, but the Zig itself was written by the model under my direction. Please review with that in mind; I'll happily iterate on any idioms, naming, or error handling that don't match the project's conventions.

## Summary

- Parsers in `src/explore.zig` hardcoded `Symbol.line_end = line_num` at the signature line, so `codedb_symbol body=true` (which calls `getSymbolBody(path, line_start, line_end, ...)`) always returned a single-line slice instead of the symbol body. Affected all 7 languages (Zig, Python, TS/JS, Rust, PHP, Go, Ruby).
- Added `computeSymbolEnds()` as a post-parse pass in `indexFileInner`, run between the streaming loop and the global lock while `content` is still in scope. Three strategies: brace-balance for C-family languages (with string/comment hygiene), indent for Python, indent + `end` snapping for Ruby. Parsers themselves are untouched.
- Naturally single-line kinds (`import`, `variable`, `constant`, `type_alias`, `macro_def`, `comment_block`) are short-circuited. Forward declarations / abstract methods (no `{` within 10 lines of the signature) stay with `line_end == line_start`.

## Commits

- `9aeca60` test(issue-224): add failing test for codedb_symbol body=true
- `8919fb8` fix(issue-224): populate Symbol.line_end via post-parse block scan

The failing test is intentionally on its own commit per the repo's issue workflow (`CLAUDE.md`).

## Test plan

- [x] `zig build test --summary all` → **333/333 passing** (324 before + 9 new).
- [x] New tests cover: Zig fn, Zig struct_def, Python def, Python class, TS function, Rust fn, Go func, PHP function, Ruby def with nested `do`/`end`, and an abstract-signature edge case that must stay single-line.
- [x] End-to-end smoke test via MCP against the built binary: `codedb_symbol name=computeSymbolEnds body=true` returns the full 36-line body of the new helper (L1939–L1974 in `src/explore.zig`), not just the signature.
- [x] Existing tests for `codedb_outline`, `codedb_read`, and `getSymbolBody` still pass (no changes to those code paths).

## Notes for reviewers

- **Design choice**: a single post-parse pass instead of refactoring each `parseXLine`. The parsers are line-streaming — giving them the rest of the file would require restructuring all 7, and any future language would need the same care. One fix location covers every current and future parser.
- **Accuracy trade-offs**: `scanBraceBlock` tracks `"..."` strings and `//` / `/* */` comments, but intentionally does not track single-quoted strings (to avoid breaking Rust lifetimes / Zig char literals) or JS template literals. A `'{'` inside a char literal or `` `${` `` inside a template string could, in theory, throw off the count — accepted as rare imprecision in exchange for a ~40-line implementation. Indent-based fallback for Python/Ruby also trades a small amount of precision (e.g., Ruby heredocs) for simplicity.
- **Stale snapshots**: existing `codedb.snapshot` files on disk were written with `line_end == line_start`. Users will need to re-index after pulling this fix to see full bodies — worth a line in the next changelog.
- **Pre-existing leak** (not in scope): while smoke-testing, I noticed `handleSymbol` in `src/mcp.zig:764` only frees the outer slice from `findAllSymbols`, leaking the duped `path`/`name`/`detail` strings of each result. It's a separate bug in code this PR doesn't touch — happy to open a follow-up issue.

### DIFF
--- a/src/explore.zig
+++ b/src/explore.zig
@@ -315,6 +315,10 @@ pub const Explorer = struct {
         }
         outline.line_count = line_num;
 
+        // Fill in Symbol.line_end for multi-line symbols (parsers only record line_start).
+        // Must happen while `content` is still in scope and before the lock is taken.
+        computeSymbolEnds(content, &outline);
+
         self.mu.lock();
         defer self.mu.unlock();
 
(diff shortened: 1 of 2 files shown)


justrach (changes requested):
I found two blocking issues that need to be addressed before this should merge.

1. Existing snapshots keep the broken `line_end` values, so the fix does not actually apply for many upgraded users. `computeSymbolEnds()` only runs during indexing, but snapshots serialize and restore `line_end` unchanged. Startup and MCP project loading both prefer `codedb.snapshot` when it matches the current git HEAD, so unchanged repos can continue serving single-line symbol bodies indefinitely until a manual reindex. Please either repair `line_end` on snapshot load / invalidate old snapshots / force recomputation somewhere on load, or otherwise make the upgrade path actually apply the fix.

2. The brace scanner still truncates bodies on valid syntax because it explicitly does not track backticks or single-quoted literals. I reproduced this locally against the PR branch: a TypeScript function containing ``const s = `}`;`` and a Zig function containing `const c: u8 = '}';` both produced `line_end = 2` instead of the closing brace line. That means `codedb_symbol body=true` is still wrong on real code, not just pathological input. Please either harden `scanBraceBlock()` for these cases or narrow the claimed language coverage and add regression tests for the remaining supported cases.

Once those are fixed, I’m happy to re-review.

@codex can you take a look as well?

ocordeiro:
@justrach
Thanks for the thorough review!

For `scanBraceBlock`, I've added single-quote and backtick tracking with escape handling, along with regression tests for Zig char literals (`'}'`) and TS template literals (`` `}` ``), in commit `a37360f`.

For the stale snapshots issue, I have a local fix ready but wanted your input before pushing. I see a few options:

1. **Bump `FORMAT_VERSION` from 2 to 3**: rejects old snapshots and forces a full reindex on next startup. Clean but heavy-handed since the user loses all cached state.
2. **Call `computeSymbolEnds` inside `insertRestoredFile`**: fixes `line_end` on load for unchanged files. Targeted, but adds recomputation to a function designed as a fast path.
3. **Post-load sweep in `loadSnapshotFast`**: after restoring all files, iterate the outlines and call `computeSymbolEnds` for each. Runs once on startup, doesn't touch `insertRestoredFile`, and corrected values get persisted when the snapshot is saved again.

Which approach do you prefer?

justrach:
Thanks for putting this together. The `#224` fix has already landed on `main`, so I am closing this PR as superseded instead of leaving a duplicate implementation open. Appreciate the careful write-up and the end-to-end validation.

PR #244, opened 2026-04-13
Allow users to set custom codedb timeout via --timeout flag

### LINKED ISSUE #243: Allow users to set custom codedb timeout
Hello,

I'm someone who really loves codedb and I use it extensively in codedb. But the biggest problem for me is the timeout, for someone who uses opencode, I sometimes spend a few moments reviewing the code, writing my own code, researching, looking at docs, or discussing with my AI on any consultations and ideas, periods where codedb isn't being used. This means I have to close opencode and then re-open the mcp server

My suggestion is to have a flag to set a timeout option like `codedb mcp . --timeout=30` for 30 minutes for e.g

Also, despite #233 saying that the process does close after the timeout... It still doesn't for me

Image

justrach:
## Summary

Adds support for custom idle timeout in the MCP server via the  flag.

## Changes

- Changed `idle_timeout_ms` from const to var in `mcp.zig`
- Added `setIdleTimeout()` function to configure timeout at runtime  
- Parse `--timeout=N` argument in main.zig for the mcp command
- Updated help text to document the new flag

## Usage

Users can now set a custom idle timeout (in minutes) like:
```
codedb mcp . --timeout=30
```

This addresses issue #243 where users needed longer timeouts for debugging sessions.

Closes #243


justrach:
Hey @destroyer22719! 

This PR implements the custom timeout feature you requested in #243. You can now use:

    codedb mcp . --timeout=30

However, I also noticed your comment about the process not actually closing after the timeout. I investigated the idle watchdog code and found a potential issue:

**The Problem:** The main MCP loop uses blocking I/O (mcpj.readLine() from stdin). When the watchdog detects idle timeout and calls stdin.close(), the blocking read may not always be interrupted properly depending on the OS and how the MCP client manages the pipe.

**Current flow:**
1. Watchdog thread detects idle timeout after 10 min (or your custom timeout)
2. Watchdog calls stdin.close() and sets shutdown = true  
3. But the main thread is blocked in readLine() and doesn't exit until it gets input

This explains why you see the process still running. The timeout is being detected, but the main thread can't exit because it's blocked waiting for input that never comes.

**Potential fix:** We'd need to either:
1. Make stdin reads non-blocking with a timeout
2. Use a select/poll approach in the main loop
3. Or have the watchdog send a signal to interrupt the read

Should I create a separate issue to track the process exit problem?

justrach:
Closing because this PR predates the 0.16 migration ([#293](https://github.com/justrach/codedb/pull/293), merged 2026-04-18) which touched core layers (storage in #283, discovery in #284, MCP/server in #285). A lot has changed and the diff no longer applies cleanly.

Please rebase onto current main and reopen — happy to review the rebased version. Thanks for the contribution!

Issue #252, opened 2026-04-13
explore: commitParsedFileOwnedOutline can leave word_index and trigram_index diverged on OOM


justrach:
## Problem

`Explorer.commitParsedFileOwnedOutline` updates `word_index`, `trigram_index`, and `outlines` in sequence with no atomic rollback. `word_index.indexFile` calls `removeFile` first (destructive) before building new entries. If `trigram_index.indexFile` subsequently fails with OOM, the indexes are left in a split state:

- `word_index`: has the **new** content (removeFile + indexFile completed)
- `trigram_index`: has the **old** content removed, new content **never inserted**
- `outlines`: restored by errdefer to the previous outline

Results: content searches via `word_index` return results for the new file version, but trigram searches miss the file entirely.

```zig
// src/explore.zig — commitParsedFileOwnedOutline
try self.word_index.indexFile(stable_path, content);    // ← removes old first, succeeds
try self.trigram_index.indexFile(stable_path, content); // ← fails OOM: word and trigram now diverged
```

## Fix

Either:
- Re-index word_index on trigram failure (errdefer that re-runs word_index.indexFile with old content)
- Or stage all mutations before committing: compute new trigrams and words without modifying indexes, then apply atomically under lock

justrach:
Fixed in release/v0.2.57 (commit 6ae58b1). commitParsedFileOwnedOutline now has an errdefer that restores word_index to prior state (re-indexes old content or calls removeFile) if trigram_index.indexFile fails.

PR #298, opened 2026-04-18
perf: CLOCK content cache for post-releaseContents search (issue #208)

### LINKED ISSUE #208: perf: CLOCK eviction cache for file contents
> **Design from [justrach/turbodb](https://github.com/justrach/turbodb)** — adapted from `src/hot_cache.zig` CLOCK eviction pattern.

## Problem

`explorer.contents` is all-or-nothing — either all file contents in memory (1.7GB on openclaw) or none (after `releaseContents`). No middle ground.

## Evidence

Head-to-head benchmark on openclaw:
- codedb: 1,729MB peak RSS
- Competitor: 41MB RSS

## Proposed Solution

Replace with a fixed-size CLOCK cache (second-chance eviction):

- Fixed 16,384 entry slots with deterministic memory footprint
- Hot files (recently searched/read) stay cached
- Cold files evicted via CLOCK sweep — reference bit cleared on first pass, evicted on second
- Cache misses fall through to `readContentForSearch` disk read

Reference: [justrach/turbodb `hot_cache.zig`](https://github.com/justrach/turbodb/blob/main/src/hot_cache.zig) — CLOCK eviction with atomic hit/miss counters, probe limit of 4, zero dynamic allocation.

### Failing test

```zig
test "issue-208: content cache evicts cold entries under pressure" {
    // Insert 100 files into cache with limit of 50
    // Verify only ~50 remain, recently accessed ones preserved
}
```

## Expected impact

| Metric | Before | After |
|--------|--------|-------|
| Peak RSS (5K files) | 1,729MB | ~100-200MB |
| Hot file latency | 0 (in HashMap) | 0 (in cache) |
| Cold file latency | N/A (released) | ~1ms (disk read) |

## Files to modify
- `src/explore.zig` — replace `contents: StringHashMap` with CLOCK cache
- `src/explore.zig` — `releaseContents`, `readContentForSearch`

justrach:
## Summary
Closes [#208](https://github.com/justrach/codedb/issues/208). Second attempt at a bounded content cache after [PR #259](https://github.com/justrach/codedb/pull/259) was reverted in [`5a5ee8a`](https://github.com/justrach/codedb/commit/5a5ee8a).

## What went wrong in PR #259
It replaced `Explorer.contents: StringHashMap` with `contents: ContentCache` (4096 slots, CLOCK eviction). Two fatal problems:

1. **`rebuildTrigrams` / `rebuildWordIndex` iterate `contents`.** Once the cache was full, subsequent puts evicted older entries silently. On a >4096-file repo, a rebuild would only re-index 4096 files; the rest disappeared from the word/trigram indexes.
2. **#252 errdefer restoration broke.** The old `getOrPut`-based flow relied on `prior_content` still living in `contents` until the errdefer could restore it. `ContentCache.put` freed the old entry immediately, so the rollback had nothing to restore.

## This PR's approach

Keep `Explorer.contents` as the primary store. Add `search_cache: ContentCache` as a separate, bounded read-through layer *only* on the disk-fallback path in [`readContentForSearch`](src/explore.zig):

```
contents.get (fast: first ~1000 files)
  → search_cache.get (warm: prior disk reads, bounded at 4096)
    → disk read → populate search_cache → return
```

### How this dodges the PR #259 bugs
| Bug | Mitigation |
|-----|------------|
| Silent index drop on rebuild | Cache is orthogonal to `contents`; no iteration path changes |
| Dangling slice on concurrent eviction | `get` returns a `dupe` (owned copy); caller handle is independent |
| Path lifetime coupling | `put` dupes path *and* data — cache lives independently of outlines |
| errdefer rollback broken | Cache is only touched inside `readContentForSearch`; no errdefer chain to violate |
| Unbounded cap mismatch | Bounded at `MAX_ENTRIES=4096`; misses fall through to disk, never silent loss |

### Where the benefit shows up
`releaseContents()` is called at end-of-scan on repos >1000 files (or when `CODEDB_LOW_MEMORY=1`). After that, `contents` is empty and every search hits disk. The new cache keeps the hot set resident and makes repeated searches to the same files ~3-10x faster than warm disk reads.

## Test plan
- [x] `zig build test` — **400/400 pass** (395 baseline + 5 new)
  - `ContentCache put/get round-trip returns owned copy`
  - `ContentCache get returns independent copy (survives eviction)`
  - `ContentCache bounds count at MAX_ENTRIES under pressure`
  - `ContentCache remove drops entry and decrements count`
  - `ContentCache clear resets state`
- [x] Benchmark on `~/openclaw` (13,777 files, 3 trials median) — no regressions: `index_ms -0.9%`, queries within ±5% noise. (The bench runs before `releaseContents`, so it hits the `contents` fast path and doesn't exercise the new layer — by design. Cache wins appear only in workloads where `releaseContents` has fired, which the benchmark doesn't currently simulate.)

### DIFF
--- a/src/explore.zig
+++ b/src/explore.zig
@@ -373,6 +373,160 @@ pub const SymbolLocation = struct {
     line_end: u32,
 };
 
+/// CLOCK (second-chance) eviction cache sitting between `readContentForSearch`
+/// and disk. Orthogonal to `Explorer.contents` — index rebuilds iterate
+/// `contents`, not this cache, so a bounded cache size never hides files from
+/// the indexer.
+///
+/// Lessons from PR #259's revert (commit 5a5ee8a):
+///   1. DO NOT replace `Explorer.contents` with a bounded cache. Functions
+///      like `rebuildTrigrams`/`rebuildWordIndex` iterate `contents` and
+///      silently drop files once capacity is exceeded.
+///   2. Return an OWNED copy from `get` so concurrent eviction by another
+///      thread can't invalidate a caller's handle.
+///   3. Dupe both path and data internally — the cache outlives the outline
+///      paths for slots it holds.
+pub const ContentCache = struct {
+    pub const MAX_ENTRIES: u32 = 4096;
+
+    const Entry = struct {
+        path: []u8 = &[_]u8{},
+        data: []u8 = &[_]u8{},
+        ref_bit: bool = false,
+        occupied: bool = false,
+    };
+
+    slots: [MAX_ENTRIES]Entry = [_]Entry{.{}} ** MAX_ENTRIES,
+    lookup: std.StringHashMap(u32),
+    hand: u32 = 0,
+    count: u32 = 0,
+    mu: cio.Mutex = .{},
+    allocator: std.mem.Allocator,
+
+    pub fn init(allocator: std.mem.Allocator) ContentCache {
+        return .{
+            .lookup = std.StringHashMap(u32).init(allocator),
+            .allocator = allocator,
+        };
+    }
+
+    pub fn deinit(self: *ContentCache) void {
+        for (&self.slots) |*slot| {
+            if (slot.occupied) {
+                self.allocator.free(slot.path);
+                self.allocator.free(slot.data);
+            }
+        }
+        self.lookup.deinit();
+    }
+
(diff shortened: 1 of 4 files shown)


justrach:
Closing — paired issue #208 closed as 'not currently a bottleneck.' The CLOCK eviction policy is a tweak on top of the existing content cache; if/when it becomes a real problem, easier to open a focused fresh PR off current main than rebase this 5-week-old branch.

Issue #311, opened 2026-04-22
codedb_remote: add wiki.codes backend alongside codegraff.com


justrach:
## Problem

`codedb_remote` today hits `codedb.codegraff.com` only. That backend is the WASM-on-Workers indexer with actions `tree | outline | search | meta`. Its sibling project [codedb-cloud / wiki.codes](https://www.wiki.codes/) is a Zig-native router on Hetzner with parquet-backed cold + hot tiers, and it offers a **superset** of useful actions — notably **`symbol`** (exact-identifier lookup) and **`policy`** (hot-pin size class) — plus more repos indexed and richer result shapes.

Agents calling codedb today have no way to reach wiki.codes. Forcing a separate MCP server for it would double the install footprint and fracture the "codedb_remote" mental model.

## Proposal

Add a `backend` parameter to the existing `codedb_remote` tool. Default preserves current behavior.

\`\`\`
codedb_remote repo="rust-lang/rust" action="symbol" query="HashMap" backend="wiki"
codedb_remote repo="vercel/next.js" action="tree" backend="wiki"
codedb_remote repo="justrach/merjs" action="search" query="handleRequest"   # still codegraff
\`\`\`

| action | codegraff (default) | wiki |
|---|---|---|
| tree    | ✓ | ✓ |
| outline | ✓ | ✓ |
| search  | ✓ | ✓ |
| meta    | ✓ | ✗ (not in router.zig) |
| symbol  | ✗ | ✓ **(new)** |
| policy  | ✗ | ✓ **(new)** |

## URL shape

- codegraff: \`https://codedb.codegraff.com/<repo>/<action>\` (existing; unchanged)
- wiki:      \`https://www.wiki.codes/api/query?slug=<slug>&endpoint=<action>&q=<query>\` via the Vercel proxy (server-side-authed, no client secrets)

Slug derivation: \`owner/repo\` → \`owner-repo\` (matches the naming wiki.codes produces on ingest: e.g. \`rust-lang/rust\` → \`rust-lang-rust\`, \`vercel/next.js\` → \`vercel-next-js\`).

## Validation

- \`backend\` ∈ \`{"codegraff","wiki"}\`, default \`"codegraff"\`
- When \`backend="wiki"\`: reject \`action="meta"\` with a clear error
- When \`backend="codegraff"\`: reject \`action="symbol"|"policy"\`
- Same repo-format + path-traversal validation as today

## Why bundle here instead of a separate MCP server

[...]

PR #314, opened 2026-04-22
codedb_remote: add 'wiki' backend (wiki.codes) — main

justrach:
Mirrors #312 onto \`main\`. Same commit, same code — I merged into \`release/0.2.579\` first without also landing on \`main\`; this PR fixes that.

Once this merges, \`main\` and \`release/0.2.579\` will both carry the \`codedb_remote backend=wiki\` capability.

See #311 for the design and #312 for the PR that already merged into release.


justrach:
Superseded — syncing main with release/0.2.579 instead so main catches up with everything, not just the wiki backend.

PR #328, opened 2026-04-26
Extend MCP idle timeout to one hour

justrach:
## Summary
- extend the MCP idle timeout from 10 minutes to 1 hour
- keep dead-client cleanup separate from idle timeout with a 1-second stdin HUP poll
- update tests and README copy for the new timeout behavior

## Verification
- zig build test
- zig build
- installed binary smoke: closing MCP stdin exited cleanly in ~285ms

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@
 | Auto-registration in Claude, Codex, Gemini, Cursor     |                                          |
 | Polling file watcher with filtered directory walker    |                                          |
 | Portable snapshot for instant MCP startup              |                                          |
-| Singleton MCP with PID lock + 10min idle timeout       |                                          |
+| Singleton MCP with PID lock + 1h idle timeout          |                                          |
 | Sensitive file blocking (.env, credentials, keys)      |                                          |
 | Codesigned + notarized macOS binaries                  |                                          |
 | SHA256 checksum verification in installer              |                                          |
--- a/src/main.zig
+++ b/src/main.zig
@@ -945,15 +945,11 @@ fn scanBg(io: std.Io, store: *Store, explorer: *Explorer, root: []const u8, allo
 }
 fn idleWatchdog(shutdown: *std.atomic.Value(bool)) void {
     const mcp = @import("mcp.zig");
+    const stdin = cio.File.stdin();
     while (!shutdown.load(.acquire)) {
-        // Sleep in 1s increments for responsive shutdown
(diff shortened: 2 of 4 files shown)


justrach:
## Speed comparison for timeout change

Compared current #328 build (`67746b5d...`, 1h idle timeout) against the immediately previous installed `0.2.579` build (`f0a40ecd...`, 10m idle timeout).

### GitHub bench-regression

Official PR bench artifact passed with all rows `OK`; no benchmark regression from the timeout/watchdog change.

| Tool | Base | Head | Delta | Status |
| --- | ---: | ---: | ---: | --- |
| `codedb_bundle` | 481005 ns | 487143 ns | +1.28% | OK |
| `codedb_outline` | 231731 ns | 238884 ns | +3.09% | OK |
| `codedb_search` | 174453 ns | 177438 ns | +1.71% | OK |
| `codedb_snapshot` | 251022 ns | 243525 ns | -2.99% | OK |
| `codedb_status` | 216322 ns | 218376 ns | +0.95% | OK |
| `codedb_symbol` | 59188 ns | 57909 ns | -2.16% | OK |
| `codedb_tree` | 87990 ns | 91863 ns | +4.40% | OK |

Full artifact had every benchmark row within the 10% / 50,000ns gate.

### Local MCP lifecycle smoke

20 local runs per binary, using the installed binaries directly:

| Measurement | Current avg / median / p95 | Previous avg / median / p95 |
| --- | ---: | ---: |
| initialize response | 40.2 / 40.6 / 46.9 ms | 41.2 / 41.5 / 42.4 ms |
| `tools/list` | 0.2 / 0.1 / 0.3 ms | 0.1 / 0.1 / 0.2 ms |
| `codedb_status` | 0.1 / 0.1 / 0.4 ms | 0.1 / 0.1 / 0.1 ms |
| normal stdin-close shutdown | 1195.5 / 1190.2 / 1278.0 ms | 1189.6 / 1188.7 / 1346.9 ms |

The lifecycle numbers are effectively unchanged for normal MCP EOF shutdown. The behavioral change in #328 is the policy: idle timeout is now 1 hour, while the dead-client watchdog poll remains independent and is bounded to 1 second.

### Remote wiki calls

Both current and previous builds successfully handled `backend=wiki` calls for `chromium` policy and `axios/axios` CVEs. Timings were network-bound and noisy, so I’m treating them as smoke checks rather than regression data.

Issue #424, opened 2026-05-05
mcp: codedb_bundle should fall through to inline args when arguments={}, plus diagnostic hints


justrach:
## Problem

Real-world MCP clients (forge agent runtimes among them) sometimes send sub-ops inside `codedb_bundle` with **empty `arguments`** and the actual args sitting inline at the op level — e.g.:

```json
{"ops":[{"tool":"codedb_outline","arguments":{},"path":"src/main.zig"}]}
```

The current dispatcher sees `"arguments"` is present, treats it as authoritative even though it's empty, and reports `missing 'path' argument` with `received keys: []` — even though `path` is right there in the op.

The diagnostic is also unhelpful in this state: `received keys: []` tells the user "you sent nothing" but doesn't suggest the inline-args fallback that would route around their broken client wrapper.

## Steps to reproduce

```bash
$ (printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"t","version":"1"}}}'; \
   sleep 1; \
   printf '%s\n' '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"codedb_bundle","arguments":{"ops":[{"tool":"codedb_outline","arguments":{},"path":"src/main.zig"}]}}}'; \
   sleep 1) | codedb mcp 2>/dev/null
```

Bundle reports:
```
error: all 1 bundle op(s) failed
--- [0] codedb_outline ---
error: missing 'path' argument
received keys: []
```

## Failing tests (3)

```zig
test "issue-424-B: bundle falls through to inline args when arguments is empty object" { ... }
test "issue-424-D: received-keys diagnostic hints at inline-args workaround when empty" { ... }
test "issue-424-A: bundle envelope errors carry the 'error:' prefix consistently" { ... }
```

All three fail on `main` at `e7f177be` (v0.2.5805).

## Expected

1. **B (medium):** when `arguments` is present but is an empty object, the dispatcher should fall through to inline-args mode rather than treating the empty object as authoritative. Sending `{tool, arguments:{}, path}` should be equivalent to `{tool, path}`.
2. **D (medium):** when `received keys:` is empty, append a hint pointing at the inline-args shape so a user with a broken client wrapper can self-recover.
[...]

PR #435, opened 2026-05-06
fix(mcp): require arguments in codedb_bundle ops schema (#434)

### LINKED ISSUE #434: mcp: codedb_bundle ops schema permits empty arguments, function-calling LLMs emit {}
## Problem

When a function-calling LLM picks `codedb_bundle`, the wire payload it emits is consistently:

```json
{
  "ops": [
    {"tool": "codedb_outline", "arguments": {}},
    {"tool": "codedb_search",  "arguments": {}}
  ],
  "project": ""
}
```

Each sub-op then fails its own arg validator with `received keys: [tool, arguments]`. The model isn't producing inline keys — both shapes are permitted by the bundle schema and the model picks the minimum-valid payload.

Root cause is the bundle's `inputSchema` itself. From `tools/list` (`src/mcp.zig:512`):

```jsonc
"ops": {
  "type": "array",
  "items": {
    "type": "object",
    "required": ["tool"],          // ← arguments NOT required
    "properties": {
      "tool": {"type": "string"},
      "arguments": {                // ← bare {} — no inner shape
        "type": "object",
        "description": "Per-call args matching that tool's inputSchema. ..."
      }
    }
  }
}
```

Two structural problems:

1. `arguments` is `{type: "object"}` with no `properties`, so `{}` is valid input.
2. `arguments` is not in the parent `required` array, so omitting or emptying it is also valid.

The natural-language description is advisory; for function-calling models the schema is authoritative.

The downstream symptom: the `#424` empty-`arguments` fallback at `src/mcp.zig:1948-1958` routes the op into inline-args mode, which uses `op_obj` (whose only keys are `tool` and `arguments`) as the args object. Each sub-tool then errors with `received keys: [tool, arguments]`.

## Failing Test

```zig
test "issue-434: codedb_bundle ops items schema requires arguments field" {
    const parsed = try std.json.parseFromSlice(std.json.Value, testing.allocator, mcp_mod.tools_list, .{});
    defer parsed.deinit();

    const tools = parsed.value.object.get("tools").?.array;
    var bundle_schema: ?std.json.Value = null;
    for (tools.items) |t| {
        const name = t.object.get("name").?.string;
        if (std.mem.eql(u8, name, "codedb_bundle")) {
            bundle_schema = t.object.get("inputSchema").?;
            break;
        }
    }
    try testing.expect(bundle_schema != null);

    const ops = bundle_schema.?.object.get("properties").?.object.get("ops").?;
    const items = ops.object.get("items").?;
    const required = items.object.get("required").?.array;

    var has_tool = false;
    var has_arguments = false;
    for (required.items) |r| {
        if (std.mem.eql(u8, r.string, "tool")) has_tool = true;
        if (std.mem.eql(u8, r.string, "arguments")) has_arguments = true;
    }
    try testing.expect(has_tool);
    try testing.expect(has_arguments);
}
```

Fails on main at `try testing.expect(has_arguments);` because the items.required array is currently `["tool"]`.

Branch: `issue-434-failing-test` (commit `d470e5b`).

## Expected

The schema should reject `arguments: {}` so well-behaved clients are forced to populate it. At minimum, `arguments` belongs in the items `required` array.

## Fix

**Stage 1 (this issue) — cheap one-line schema change in `src/mcp.zig:512`:**

```jsonc
"items": {
  "type": "object",
  "required": ["tool", "arguments"],   // was: ["tool"]
  ...
}
```

Doesn't constrain the contents, but kills the empty-`{}` failure path immediately. Models that previously emitted `arguments: {}` will populate something, and the inline-shape fallback at `mcp.zig:1948` stops being load-bearing for conformant clients.

**Stage 2 (follow-up) — discriminated union over `tool`:**

```jsonc
"items": {
  "type": "object",
  "required": ["tool", "arguments"],
  "oneOf": [
    { "properties": { "tool": {"const": "codedb_outline"}, "arguments": <codedb_outline.inputSchema> } },
    { "properties": { "tool": {"const": "codedb_search"},  "arguments": <codedb_search.inputSchema>  } }
    // ... one branch per codedb_* sub-tool
  ]
}
```

Each branch pins `tool` to a const and `arguments` to that sub-tool's actual `inputSchema`. The model is then bound by the schema to populate `arguments` correctly for whichever tool it picked.

Caveats for Stage 2:
- The `tools_list` literal at `src/mcp.zig:497-520` is hand-rolled and would need to become a builder so the per-sub-tool schemas can be composed at startup instead of duplicated by hand.
- Schema payload size roughly doubles (~20 sub-tools × their inputSchemas inlined).
- `oneOf` enforcement varies across providers — Anthropic accepts it but isn't strict the way OpenAI structured-output is, and some MCP clients pass schemas straight through. The `#424` inline-shape fallback should stay as a backstop regardless.

Repro: any MCP client whose LLM emits structured tool calls hits this. Confirmed in graff 0.1.42-dev and graff 0.1.5 against codedb operating on `~/nanobrew`.

justrach:
## Summary

- Adds `"arguments"` to the `codedb_bundle` ops items `required` array so function-calling LLMs are forced to populate it instead of emitting `{tool, arguments: {}}`.
- Includes a failing test (`issue-434`) that introspects the `tools_list` schema and asserts `arguments` is required for ops items.

Fixes #434.

## Background

Prior schema had ops items as `required: ["tool"]` with `arguments` as a bare `{type: "object"}`. Function-calling models read the schema as authoritative and emitted the minimum-valid payload — `{tool: "...", arguments: {}}` — which then misrouted through the `#424` inline-args fallback (since the empty `{}` triggers fallthrough at `mcp.zig:1948`). The op object's only keys are `tool` and `arguments`, so each sub-tool errored with `received keys: [tool, arguments]`.

This is **Stage 1** (one-line schema change). Stage 2 — discriminated `oneOf` over `tool` binding `arguments` to each sub-tool's actual `inputSchema` — is a follow-up issue: it requires turning the hand-rolled `tools_list` literal at `src/mcp.zig:497-520` into a builder so per-sub-tool schemas can be composed at startup.

The runtime inline-args fallback at `mcp.zig:1948` stays as a backstop for non-conformant clients.

## Commits

1. `d470e5b` — adds failing test, exposes `tools_list` as `pub` for the test to introspect.
2. `7fb1e87` — applies the schema fix; updates the description text in the same line to reflect that `arguments` is now required.

## Test plan

- [x] `zig build test` — 507/507 pass with the fix; was 506/507 (issue-434 failing) without it.
- [ ] Confirm against a real MCP client (graff or Claude) that bundle calls now arrive with populated `arguments`.


justrach:
Superseded by #439 (release v0.2.5808). All commits from this branch landed in 907ac96 on main.

PR #436, opened 2026-05-06
feat(explore): opt-in rerank-trace logger for offline tuning

justrach:
## Summary

Adds an opt-in JSONL logger that captures every `searchContent` call so we can collect data for offline rerank tuning experiments. Pure observation — does not change ranking behavior.

Enable via `.codedbrc`:

```
rerank_trace = true
```

When set, each search appends one line to `<data_dir>/rerank-traces.jsonl`:

```json
{"ts":1778294611234,"query":"widgetX","results":[{"path":"src/widgetX.zig","line":1,"score":21.0000},{"path":"src/unrelated.zig","line":1,"score":1.0000}]}
```

Off by default. Capped: 256 byte queries, 50 results per line, 10 MB file (rotates by truncate-clobber). All I/O errors are swallowed — logging never breaks a search.

## Why

Today the multi-signal rerank in [`rerankSignalScore`](src/explore.zig:1748) uses 7 hand-tuned constants (+5 sym-def, +15/+8 basename, +6 path-seg, ×0.6 tests, ×0.4 vendor, ×0.5 doc). Different repos want different priors (a tests-heavy monorepo wants the ×0.6 test penalty softened; a vendored-deps repo wants ×0.4 cranked harder). Before deciding whether online learning-to-rank from agent traces is worth building, we want to look at the raw data first. This PR is just the v0 logger to enable that.

## Commits

- **54b6b72** — feat: the logger itself, opt-in via `.codedbrc`
- **d4391db** — fix: caught by end-to-end binary verification — pre-fix, single-result queries logged `score:0.0` because the scoring loop was guarded behind `len > 1`. Lifted scoring out of the guard, kept the sort inside.

## Implementation notes

- New field `Config.rerank_trace: bool` parsed from `.codedbrc` (`src/config.zig`)
- New field `Explorer.rerank_trace_path: ?[]const u8` (borrowed slice; caller-owned for the Explorer's lifetime). Set from `main.zig` to `<data_dir>/rerank-traces.jsonl` when the flag is on.
- `appendRerankTrace` called from `rerankAndFinalize` — runs inside the existing read lock; writes are protected by an advisory file lock for cross-process safety; uses a 16 KB stack buffer with conservative per-entry budgeting so partial entries never produce invalid JSON.
- `writeJsonEscaped` free helper for minimal JSON escaping (`"`, `\`, control chars).

## Test plan

- [x] `zig build test` — 511/511 pass (4 new tests; main was 506)
- [x] Unit tests cover: opt-in writes one line, opt-out writes nothing, size cap clobbers correctly, single-result entries get non-zero scores, config parses `true`/`false`/malformed
- [x] **End-to-end binary verification (Sonnet 4.6 agent run twice):**
  - Build clean
  - Default off → no file created
  - Opt-in via `.codedbrc` → file appears at `~/.codedb/projects/<hash>/rerank-traces.jsonl`
  - All 9 lines (varied queries) parse as valid JSON via `python3 json.loads`
  - JSON escaping for `"` and `\` round-trips correctly
  - Multi-result queries: `widget.zig:1 (17.0) > widget.zig:2 (16.0) > other.zig:1 (1.0) > other.zig:2 (1.0)` — sort order preserved
  - Single-result query post-fix: `score: 1.0` (pre-fix would have been `0.0`)

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -17,6 +17,10 @@ pub const Config = struct {
     max_versions: usize = 100,
     /// Cap on files kept in the Explorer's in-memory content cache. Default 1000.
     max_cached: u32 = 1000,
+    /// When true, append one JSON line per searchContent invocation to
+    /// <data_dir>/rerank-traces.jsonl. v0 logger for offline rerank-tuning
+    /// experiments. Off by default — opt in via .codedbrc.
+    rerank_trace: bool = false,
 
     pub const default: Config = .{};
 
@@ -37,6 +41,8 @@ pub const Config = struct {
             } else if (std.mem.eql(u8, key, "max_cached")) {
                 cfg.max_cached = std.fmt.parseInt(u32, val, 10) catch return error.InvalidMaxCached;
                 if (cfg.max_cached == 0) return error.InvalidMaxCached;
+            } else if (std.mem.eql(u8, key, "rerank_trace")) {
+                cfg.rerank_trace = parseBool(val) catch return error.InvalidRerankTrace;
             }
         }
         return cfg;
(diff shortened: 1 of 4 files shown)


justrach:
Superseded by #439 (release v0.2.5808). All commits from this branch landed in 907ac96 on main.

PR #463, opened 2026-05-20
fix(explore): short-circuit Tier 5 full scan on definitively-negative queries (#462)

### LINKED ISSUE #462: explore: negative-query Tier 5 short-circuits missing — 113ms p50 on react corpus
## Problem

`Explorer.searchContent` has a Tier 5 full-outline-scan fallback that fires whenever every earlier tier returns zero results. For a definitively-negative query — one whose trigrams don't exist anywhere in the index — Tiers 0–4 all no-op and Tier 5 then scans every indexed file to find nothing.

On the React corpus (6,619 files, ~26.5 MB), a query like `xyzzy_react_does_not_exist` measures at **113ms p50** (p99 ~290ms) in `benchmarks/search-shootout`. The work is entirely wasted: the trigram index has already proven no file matches.

## Failing test

```zig
test "issue-negq: negative-query search short-circuits Tier 5 full scan" {
    defer arena.deinit();
    var explorer = Explorer.init(arena.allocator());

    // Index enough files that Tier 5 would be observably wasteful if it ran.
    var i: usize = 0;
    while (i < 50) : (i += 1) {
        var buf: [32]u8 = undefined;
        const path = try std.fmt.bufPrint(&buf, "file_{d}.zig", .{i});
        try explorer.indexFile(path, "fn process() void { _ = thing; }\n");
    }

    // 'zzqqxxnopematch' — trigrams none of which appear in any indexed file.
    const results = try explorer.searchContent("zzqqxxnopematch", testing.allocator, 10);
    defer { ... }

    try testing.expectEqual(@as(usize, 0), results.len);
    // The fix: Tier 5 must NOT fire when the trigram index has already
    // ruled out a match. On main this expectation fails (count == 1).
    try testing.expectEqual(@as(u64, 0), explorer.search_tier5_count);
}
```

The test adds a test-only counter `Explorer.search_tier5_count` that increments each time the Tier 5 block executes. Before the fix the counter goes to 1 and the assertion fails.

## Expected

When `trigram_index.candidates(query)` returns a non-null but empty candidate set with `query.len >= 3`, every trigram-indexed file is provably free of the query. The only files that could still contain a match are `skip_trigram_files`, which Tier 3 already scanned. Tier 5 should be skipped.

Negative-query p50 should drop from ~113ms to <1ms on the React corpus.

## Fix

Six-line guard added to the Tier 5 block in `searchContent`:

```zig
const trigram_ruled_out = if (candidate_paths) |cp|
    (cp.len == 0 and query.len >= 3)
else
    false;
if (result_list.items.len == 0 and !trigram_ruled_out) {
    // Tier 5 full scan
}
```

Measured after fix on the same React corpus: `xyzzy_react_does_not_exist` **113.39ms → 0.29ms p50 (390× speedup)**.

Fix and the failing test are on branch `issue-negq-shortcircuit-failing-test`. PR incoming.

Discovered via `benchmarks/search-shootout/` (also added on the same branch).

justrach:
## Summary

- Adds a 6-line guard to `searchContent`'s Tier 5 full-outline-scan fallback so it doesn't fire when the trigram index has already proven no file can contain the query.
- Adds `benchmarks/search-shootout/` — the reproducible harness that surfaced this slow path (codedb vs SQLite FTS5 trigram/unicode61 vs lean-ctx on facebook/react).
- Closes #462.

## Before / after (React corpus, 6,619 files, p50)

| Query | Before | After | Speedup |
|---|---|---|---|
| `xyzzy_react_does_not_exist` | 113.39 ms | **0.29 ms** | **390×** |

## How the bug works

`searchContent` has 6 tiers (0, 0.5, 1, 2, 3, 4, 5). Tier 5 is the full-outline-scan fallback, gated on `result_list.items.len == 0`. For a query whose trigrams don't exist in the index at all, the candidate set returned by `trigram_index.candidates(query)` is a non-null empty slice — definitively saying "no trigram-indexed file matches." Tiers 0, 0.5, 1, 2, 4 all no-op. Tier 3 scans `skip_trigram_files`. Then Tier 5 fires and scans every file in `outlines` to find nothing.

## How the fix works

When `trigram_index.candidates(query)` returns a non-null but empty candidate set with `query.len >= 3`, every trigram-indexed file is provably free of the query. The only files that could still contain a match are `skip_trigram_files`, which Tier 3 already scanned. So Tier 5 can be skipped.

```zig
const trigram_ruled_out = if (candidate_paths) |cp|
    (cp.len == 0 and query.len >= 3)
else
    false;
if (result_list.items.len == 0 and !trigram_ruled_out) {
    self.search_tier5_count += 1;
    // ... full scan ...
}
```

## Test approach

Time-based perf tests are flaky, so the failing test uses a test-only counter (`Explorer.search_tier5_count`) that increments when the Tier 5 block runs. The test asserts the counter stays at 0 for a synthetic negative query. Per project convention, the failing test landed in a separate commit before the fix.

## Test plan

- [x] `zig build test --test-filter "issue-negq"` — passes after fix, fails on main
- [x] `python3 benchmarks/search-shootout/shootout.py --corpus ~/codedb-bench/react` — measured 113ms → 0.29ms
- [ ] Reviewer: sanity-check the candidate_paths null vs empty semantics in `trigram_index.candidates`
- [ ] Reviewer: confirm `skip_trigram_files` always contains every file in `outlines` that isn't in `trigram_index` (so Tier 5 is genuinely redundant after Tier 3)

## Out of scope

The `flushPassiveEffects`-shaped queries also show outlier p50 latencies (~167ms) in the bench but go through a different code path (Tier 1, not Tier 5). Filed separately if reproducible after this lands.

### DIFF
--- a/benchmarks/search-shootout/.gitignore
+++ b/benchmarks/search-shootout/.gitignore
@@ -1,0 +1,4 @@
+*.db
+*.db-wal
+*.db-shm
+__pycache__/
--- a/benchmarks/search-shootout/README.md
+++ b/benchmarks/search-shootout/README.md
@@ -1,0 +1,83 @@
+# search-shootout
+
+Compares full-text search backends for code intelligence on a real-world
+JS/TS corpus (facebook/react):
+
+| Backend | What it is |
+|---|---|
+| `codedb` | This repo. Custom Zig trigram + inverted word index, served over MCP. |
+| `fts5_trigram` | SQLite FTS5 with the `trigram` tokenizer (added in 3.34). Pure-substring matching, similar shape to codedb's trigram. |
+| `fts5_unicode61` | SQLite FTS5 with the default word-boundary tokenizer + BM25 ranking. The "lean-ctx-style" backend. |
+| `lean-ctx` | The yvgude/lean-ctx CLI (`lean-ctx index build` + `lean-ctx grep`). |
+
+## Reproducing
+
+1. Build codedb: `zig build -Doptimize=ReleaseFast`
+2. (Optional) Install lean-ctx: `cargo install lean-ctx`
+3. Clone a corpus **outside `/tmp`** — codedb refuses to index temporary roots:
+   ```
+   mkdir -p ~/codedb-bench
+   git clone --depth 1 https://github.com/facebook/react ~/codedb-bench/react
+   ```
+4. Run:
+   ```
+   python3 shootout.py --corpus ~/codedb-bench/react \
+                       --out results/react-$(date +%Y-%m-%d).md \
+                       --clean-codedb
+   ```
+
+Flags:
+
+- `--iters N` — warm iterations per query (default 20)
+- `--skip-codedb`, `--skip-fts5`, `--skip-leanctx` — limit backends
+- `--clean-codedb` — wipe matching codedb snapshot before indexing (forces cold build)
+
+## What it measures
+
+**Build phase:** wall-clock time to build the index from scratch + final on-disk
+size. Run once per backend per corpus.
+
+**Query phase:** for each query in `queries.json`, warm the backend with one
+call, then time `--iters` calls. Reports p50 and p99 latency plus the result
+count.
+
+> ⚠️  Result counts are **not directly comparable** across backends:
+> - codedb counts matching lines (and caps display at 50)
+> - FTS5 counts matching files
+> - lean-ctx counts matches reported in its "N matches in M files" header
+>
+> Use these as a **recall sanity check** (zero vs non-zero) and use
+> **latency** for performance comparison.
+
+## Calibration: how each backend is invoked
+
+- **codedb** runs as an MCP server over stdio (`codedb <root> mcp`). The first
+  query warms a long-lived process; subsequent queries are pure RPC.
+- **FTS5** uses a persistent SQLite connection from Python.
+- **lean-ctx** is invoked as `lean-ctx grep <q>` with `cwd=<corpus>` per
+  query. Each invocation pays ~700ms of CLI binary startup even though
+  lean-ctx uses a daemon for the actual search work. This is honest "what it
+  costs to call from a script" — not what an MCP-resident lean-ctx would feel
+  like.
+
+## Query set
+
+`queries.json` covers a spectrum of code-search workloads on a React corpus:
+
+- common-identifier — high-frequency, lots of result merging (`useState`)
+- camelcase-identifier — full identifier (`forwardRef`, `createElement`)
+- substring-identifier — substring of bigger names (`Fiber`, `Lane`, `Suspense`)
+- rare-camelcase — low-frequency named symbol (`flushPassiveEffects`)
+- short-trigram-exact — 3-char query, exercises the trigram lower bound (`set`)
+- lang-keyword — high-frequency stress test (`function`)
+- negative — should return 0 on all backends (`xyzzy_react_does_not_exist`)
+
+## Output
+
+Console: one line per query showing each backend's p50/p99 and hit count.
+Markdown: written to `--out` path, includes corpus stats, build table, and
+per-query latency table.
+
+Reports for past runs live under `results/`. A separate `RESULTS.md` at the
+top of this folder summarizes findings across runs and includes the agentic
+traversal eval (a Sonnet 4.6 sub-agent doing the same task on each backend).
--- a/benchmarks/search-shootout/RESULTS.md
+++ b/benchmarks/search-shootout/RESULTS.md
@@ -521,3 +521,87 @@ technically available.
 
 All four are opt-in. Default behavior unchanged.
 
+## 11. Engine-vs-engine — direct comparison (no MCP, no formatting)
+
+Earlier sections compared codedb to FTS5 with codedb running through MCP
+(stdio JSON-RPC) and FTS5 running through a persistent Python SQLite
+connection. That comparison was fair as a "what does the agent feel"
+benchmark, but it conflated codedb's pure engine work with its
+~0.08–0.26 ms MCP roundtrip floor and ~1 ms of content-reading per result.
+
+A new subcommand `codedb bench-engine` exposes the engine directly:
+loads the snapshot, calls the explorer in a tight loop, reports timing
+in nanoseconds. Four ops: `word`, `word-fmt`, `search`, `search-fmt`.
+
+200 iter, warm, React corpus:
+
+### Word index direct lookup (`searchWord`)
+
+| Query | hits | codedb engine | FTS5 trigram | codedb is |
+|---|---|---|---|---|
+| `useState` | 2,689 | **8 µs** | 814 µs | **102× faster** |
+| `forwardRef` | 466 | **1 µs** | 179 µs | **179× faster** |
+| `Fiber` | 6,304 | **25 µs** | 132 µs | **5× faster** |
+| `flushPassiveEffects` | 23 | **~0 µs** | 210 µs | **>200× faster** |
+| `xyzzy_react_does_not_exist` | 0 | **~0 µs** | 201 µs | **>200× faster** |
+
+**The codedb engine is 5×–200× faster than SQLite FTS5 trigram at
+word-index lookup.** This is what the agent-level comparison was hiding:
+codedb's pure inverted-word-index path is dramatically faster than FTS5;
+the per-call latency we measured at the MCP level was overhead + content
+reads, not engine cost.
+
+### Full search (`searchContent` — includes per-result content reads)
+
+| Query | codedb engine | FTS5 trigram | ratio |
+|---|---|---|---|
+| `useState` (50 results) | 1.52 ms | 0.81 ms | fts5 1.86× faster |
+| `forwardRef` | 0.17 ms | 0.18 ms | tied |
+| `Fiber` | 0.22 ms | 0.13 ms | fts5 1.7× faster |
+| `flushPassiveEffects` | 0.04 ms | 0.21 ms | **codedb 5× faster** |
+| `xyzzy_react_does_not_exist` | **0.002 ms** | 0.20 ms | **codedb 100× faster** |
+
+Reading line text from 50 files (codedb returns `path:line:line_text`)
+takes ~1 ms; FTS5's response is `path` only, so it doesn't pay this
+cost. For common identifiers FTS5 wins by ~2×; for rare identifiers and
+negative queries (with the Tier 5 short-circuit), codedb dominates.
+
+### Format-only overhead
+
+`word` vs `word-fmt` (raw search vs search + format-to-buffer):
+
+| Query | word | word-fmt | added by format |
+|---|---|---|---|
+| useState (2,689 hits) | 8 µs | 60 µs | 52 µs (~20 ns/hit) |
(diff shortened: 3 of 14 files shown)


justrach:
## Extended bench: lean-ctx MCP head-to-head

Just pushed `87db0fc` which adds an MCP-stdio client for lean-ctx so the comparison is fair — both backends as persistent warm processes, no per-call CLI binary startup penalty.

**MCP-vs-MCP on the React corpus (25 iter, warm):**

| Query | codedb p50 | lean-ctx MCP p50 | codedb wins by |
|---|---|---|---|
| `useState` | 3.74 ms | 40.04 ms | **10.7×** |
| `forwardRef` | 0.29 ms | 44.43 ms | **153×** |
| `Fiber` | 0.66 ms | 106 ms | **160×** |
| `flushPassiveEffects` | 0.46 ms | 160 ms | **348×** |
| `function` (stress, ~5k matches) | 15.76 ms | 44.10 ms | **2.8×** |
| `xyzzy_react_does_not_exist` (post-fix) | 0.13 ms | 182 ms | **1400×** |

codedb is faster on every query in the 15-query set. The previous (CLI-mode) numbers were significantly penalizing lean-ctx by including Rust binary startup; this version is honest.

## Tier 0 attribution

Added an MCP-roundtrip-floor baseline so per-query latency can be attributed. For `useState`:
- 0.08 ms MCP roundtrip
- 1.73 ms word-index hit collection + path:line formatting
- 1.06 ms content reads + line extraction
- = 2.87 ms p50 total

## Bonus finding worth a separate issue

p99 across **every** codedb MCP tool — including the supposedly O(1) `codedb_find` — was in the 300–900 ms range over 50 iterations. That's not engine cost; it's periodic ~500 ms+ spikes from something (snapshot persistence? watcher polling? arena cleanup?). The bench is reproducible — worth filing.

See `benchmarks/search-shootout/RESULTS.md` sections 7 and 8 for the full writeup.

justrach:
## Multi-task agentic eval — direct test of lean-ctx's "99% token savings"

Pushed `930f54e`. Ran 12 Sonnet 4.6 sub-agents in parallel (4 React-internals exploration tasks × 3 backends), each restricted to its backend's CLI only.

**All 12 agents reached the correct answer.** So the differentiator is purely efficiency:

| Backend | Total tool calls | Total wall sec | Total tokens (4 tasks) | Avg per task |
|---|---|---|---|---|
| **fts5_trigram** | **33** | **198 s** | **63,559** | 8.2 calls, 49 s, 15.9k tok |
| codedb | 42 | 235 s | 78,424 | 10.5 calls, 59 s, 19.6k tok |
| lean-ctx | 53 | 396 s | 78,604 | 13.2 calls, 99 s, 19.7k tok |

### The headline (which directly contradicts lean-ctx's marketing)

lean-ctx pitches "up to 99% token savings." Across these agent tasks:
- lean-ctx used **1.00× the tokens of codedb** — essentially identical total spend
- lean-ctx used **1.24× the tokens of fts5_trigram** — 24% MORE than the plainest backend
- lean-ctx took **1.69× the wall time of codedb** to complete the same tasks

Per-output compression is real, but the compression is lossy enough that the agent has to do follow-up probes to reconstruct what it needs. Per-call savings get spent on extra calls. Net token bill: no improvement vs codedb, worse than raw sqlite3.

Quote from the T1 lean-ctx agent, verbatim:
> "The build uses minified symbol aliases (α1) for cross-module exports; α1 in ReactFiberHooks.js maps to scheduleUpdateOnFiber from ReactFiberWorkLoop.js."

That's the compression actively producing work for the agent to undo.

### Where codedb lands

Middle ground. Token-equivalent to lean-ctx, 40% faster at task completion, while offering structural features (outline/find/deps) neither alternative has.

Full per-task table + caveats in `benchmarks/search-shootout/RESULTS.md` §9.

### Caveats
- 4 tasks is a probe, not a benchmark suite.
- All three backends used CLI access only. lean-ctx MCP-mode (used in §7's latency table) wasn't exposed to the agents — that's a separate experiment.

justrach:
## Tried to make codedb cheaper — proved the compression trap on our own tool

After the §9 finding that lean-ctx's compression hurt agents, I tried to add the same kind of compression to codedb and see if it would land differently. It didn't.

### Flags added (all opt-in, default unchanged)

- `CODEDB_MCP_LEAN=1` — strips ANSI summary + guidance blocks from MCP responses (~5% per call)
- `CODEDB_QUIET=1` — CLI equivalent
- `paths_only=true` / `--paths-only` — emits `path:line` without line text (~20% per call)
- `TextContent.annotations.audience` — spec-canonical per MCP, lets clients strip user-only blocks

### Re-ran the agentic eval with codedb agents told the new flags existed

| Task | codedb default | codedb LEAN+--paths-only |
|---|---|---|
| T1 setState trace | 14 / 108s / 21,010 tok | 27 / 226s / 35,504 tok |
| T2 Snapshot sites | 10 / 55s / 18,758 tok | 7 / 46s / 20,123 tok |
| T3 compare 2 fns | 8 / 28s / 16,058 tok | 10 / 52s / 17,795 tok |
| **Total** | 32 / 191s / 55,826 tok | 44 / 324s / 73,422 tok |
| **Ratio** | 1.00× | **1.38× calls, 1.70× wall, 1.32× tokens** |

**Lean variant used 32% MORE tokens than default codedb.** Same compression-trap we documented for lean-ctx — except this time on our own tool.

### Lesson

Codedb's default rich response (path:line:line_text) is the right design for agents. The flags ship as **opt-in for non-agent use** and are explicitly marked "NOT for agents" in the schema. Adding compression options to an agent-facing tool gives the LLM a foot-gun: decision overhead + follow-up calls eat the per-call savings.

Full writeup: `benchmarks/search-shootout/RESULTS.md` §10.

### Net state of the PR

5 distinct changes:
1. Negative-query Tier 5 short-circuit (the original fix, 113ms → 0.29ms)
2. Full reproducible bench harness
3. lean-ctx MCP head-to-head (codedb 10–1400× faster)
4. Multi-task agentic eval (codedb token-equivalent to lean-ctx, 40% faster wall-time)
5. **This:** opt-in compression flags + the honest finding that they hurt agents

justrach:
## Engine-vs-engine direct measurement — codedb is 5–200× faster than FTS5 trigram

Added a `codedb bench-engine <op> <query>` subcommand that bypasses MCP/CLI envelope and measures the explorer paths directly. This lets us compare to FTS5's raw sqlite3 interface fairly.

Re-ran on React corpus (200 iter, warm):

### Word-index lookup (codedb_word path)

| Query | hits | codedb engine | FTS5 trigram | codedb wins by |
|---|---|---|---|---|
| `useState` | 2,689 | **8 µs** | 814 µs | **102×** |
| `forwardRef` | 466 | **1 µs** | 179 µs | **179×** |
| `Fiber` | 6,304 | **25 µs** | 132 µs | **5×** |
| `flushPassiveEffects` | 23 | **~0 µs** | 210 µs | **>200×** |
| `xyzzy_react_does_not_exist` | 0 | **~0 µs** | 201 µs | **>200×** |

### Full search (codedb_search path — includes per-result content reads)

| Query | codedb engine | FTS5 trigram | ratio |
|---|---|---|---|
| `useState` | 1.52 ms | 0.81 ms | fts5 1.86× (50 file opens cost) |
| `Fiber` | 0.22 ms | 0.13 ms | fts5 1.7× |
| `forwardRef` | 0.17 ms | 0.18 ms | tied |
| `flushPassiveEffects` | 0.04 ms | 0.21 ms | **codedb 5×** |
| `xyzzy_react_does_not_exist` | **0.002 ms** | 0.20 ms | **codedb 100×** (negq-fix working!) |

### What the earlier MCP-vs-MCP comparison was hiding

The pure codedb engine is dramatically faster than FTS5 at word-index lookup. The "codedb 3-15 ms vs fts5 0.01-2 ms" numbers from §2 were dominated by **MCP roundtrip floor + per-result content reads**, not engine cost. The search-side gap (codedb 1.5 ms vs fts5 0.8 ms for useState) is the cost of returning line-text for 50 results — FTS5 returns paths only, which is why it's faster for the search op.

### Profile finding: format-to-buffer is negligible

`word` vs `word-fmt` shows formatting adds ~20 ns/hit. The 1.7 ms MCP-layer cost I was investigating isn't in the serializer — it's MCP envelope construction + content reads. No serializer optimization needed.

Full writeup in `benchmarks/search-shootout/RESULTS.md` §11. Reproducible:

```bash
zig build -Doptimize=ReleaseFast
./zig-out/bin/codedb /path/to/corpus bench-engine word useState 200
```

justrach:
## Second-biggest fix this PR surfaced: telemetry tail latency

Precision rerun (500 iter × 3 sessions, min/p50/p95/p99 + normalized files-list jaccard) showed codedb's MCP p99 was consistently 250–400 ms across **every** tool — even O(1) ones like `codedb_find`. That ruled out engine cost.

Root cause traced to `Telemetry.record()` calling `syncToCloud()` every 10 events, which shells out to `curl --max-time 5` on the same thread as the tool response. ~10% of every codedb MCP call was being blocked by a synchronous network call.

Fix (commit `4369d7d`): removed the in-line cloud-sync trigger. Cloud sync now happens only on `Telemetry.deinit()` shutdown. Local WAL flush every 3 events still happens (fast, <1 ms).

### codedb p99 before/after

| Query | Before p99 | After p99 | Improvement |
|---|---|---|---|
| `useState` | 363 ms | **1.81 ms** | 200× |
| `forwardRef` | 368 ms | **0.29 ms** | 1,269× |
| `Fiber` | 399 ms | **0.37 ms** | 1,078× |
| `flushPassiveEffects` | 349 ms | **0.12 ms** | 2,908× |
| `xyzzy_react_does_not_exist` | 347 ms | **0.07 ms** | 4,957× |

**Average p99 improvement across 15 queries: ~1,200×.**

### codedb vs everything else (p99, post-fix)

| Query | codedb | fts5_tri | fts5_uni | lean-ctx MCP |
|---|---|---|---|---|
| `useState` | **1.81** | 2.22 | 0.02 | 65.95 |
| `forwardRef` | 0.29 | 0.21 | 0.01 | 69.01 |
| `flushPassiveEffects` | 0.12 | 0.24 | 0.01 | 198.92 |
| `xyzzy_react_does_not_exist` | **0.07** | 0.24 | 0.04 | 242.25 |

codedb's p99 is now competitive with FTS5 trigram (within 2-3×) and 10–1400× faster than lean-ctx MCP at every percentile, not just median.

### Final state of the PR

8 commits, 5 PR comments. Substantive findings:
1. **Negative-query Tier 5 short-circuit** — 113 ms → 0.29 ms (390×)
2. **Telemetry in-line `syncToCloud` removal** — average p99 1,200× improvement
3. Full reproducible bench harness (codedb vs FTS5 vs lean-ctx, MCP + agentic)
4. **Engine-direct numbers**: codedb word index is 5–200× faster than FTS5 trigram
5. **lean-ctx MCP head-to-head**: codedb wins every query, 10–1400×
6. **Multi-task agentic eval (N=12)**: all backends correct, codedb 40% faster wall time, token-equivalent to lean-ctx
7. **Self-experiment** showing compression on our own tool burned 32% MORE agent tokens (the cleanest evidence we have for the "compression hurts agents" thesis)
8. **Opt-in lean-mode env vars + paths-only** — ship, but explicitly marked "NOT for agents" in the schema

Standalone reproducible bench at

Issue #508, opened 2026-05-28
codedb_remote returns api.wiki.codes HTTP 530 / code 1033 for public repos


ahndohun:
## Summary

`codedb_remote` is advertised as the official path for querying indexed public GitHub repositories via `api.wiki.codes`, but it currently returns HTTP 530 / error code 1033 for multiple public repositories, including the `codedb` repository itself.

I am a paid CodeGraff Pro user relying on CodeGraff/codedb as the primary Codex navigation path. This failure makes official-repo inspection unreliable and forces agents to fall back to other web/GitHub tools.

## Environment

- `codedb --version`: `codedb 0.2.5820`
- CodeGraff Pro: `muonry status` reports `Codegraff Pro — license valid`
- Client surface: Codex Desktop MCP `mcp__codedb__codedb_remote`
- Date observed: 2026-05-28

## Reproduction

Run either of these remote reads:

```text
codedb_remote repo="justrach/codedb" action="read" path="README.md" lines="1-20"
codedb_remote repo="openai/codex" action="read" path="README.md" lines="1-20"
```

Observed output for `justrach/codedb`:

```text
✗ • remote — api.wiki.codes HTTP 530 for justrach-codedb/read — error code: 1033
error: api.wiki.codes HTTP 530 for justrach-codedb/read — error code: 1033
```

Observed output for `openai/codex`:

```text
✗ • remote — api.wiki.codes HTTP 530 for openai-codex/read — error code: 1033
error: api.wiki.codes HTTP 530 for openai-codex/read — error code: 1033
```

The same pattern was also observed against other public repos during the same session, including `Yeachan-Heo/oh-my-codex` and `openclaw/openclaw`.

## Expected behavior

For indexed public repos, `codedb_remote` should return the requested content or a clear non-530 explanation such as `repo not indexed`, `path not indexed`, `rate limited`, or `service unavailable with retry-after`.

## Actual behavior

The tool returns HTTP 530 / code 1033. There is no actionable remediation, retry guidance, or service-status indication.

## Impact

This blocks the documented CodeGraff-first workflow for official upstream repo inspection. It also makes it hard to distinguish a repo/content issue from an infrastructure outage.

## Requested resolution

Please provide an official fix, public status explanation, or supported fallback guidance for `codedb_remote`. If the remote public-repo feature cannot be made reliable for paid users, please clarify whether this is covered by the CodeGraff refund policy.

justrach:
[...]

Issue #507, opened 2026-05-28
search misses existing file content after snapshot rebuild while tree/read show the file


ahndohun:
## Summary

I am a paid CodeGraff Pro user and I am seeing inconsistent local `codedb` search behavior on `codedb 0.2.5820`.

After rebuilding the local snapshot for `/Users/ahndohun/Documents/doran-orchestrator`, `codedb tree` shows the target files, and `zigread` confirms the text exists in the file, but `codedb search` returns no results for that exact text.

This makes `codedb search` unsafe as the primary navigation path for Codex/MCP workflows because a negative result is not reliable.

## Environment

- Platform: macOS darwin arm64
- `codedb --version`: `codedb 0.2.5820`
- CodeGraff Pro: `muonry status` reports `Codegraff Pro — license valid`
- Codex MCP registration:

```text
codedb
  enabled: true
  transport: stdio
  command: /Users/ahndohun/bin/codedb
  args: /Users/ahndohun/Documents/doran-orchestrator mcp
  cwd: -
  env: -
  startup_timeout_sec: 30
```

## Reproduction evidence

From `/Users/ahndohun/Documents/doran-orchestrator`:

```sh
/Users/ahndohun/bin/codedb /Users/ahndohun/Documents/doran-orchestrator tree
```

The tree includes:

```text
bin/
  doran-openclaw-orchestrator  unknown  63L  0 sym
openclaw/
  maintenance.sh  shell  405L  46 sym
```

The text exists in the file:

```sh
/Users/ahndohun/bin/zigread -L 41-47 bin/doran-openclaw-orchestrator
```

Output:

```text
41 | if [[ "${OPENCLAW_ORCHESTRATOR_INJECT_POLICY:-1}" == "1" ]]; then
42 |   policy_context="$(cat <<'POLICY'
43 | Doran Orchestrator operating contract:
44 | - AIHero / Matt Pocock skills from AGENTS.md are the first-principle development workflow.
45 | - Use oh-my-codex / OMX only as a secondary runtime and orchestration layer: roles, hooks, HUD/state, durable goal support, and coordinated execution.
46 | - When AIHero and OMX overlap, prefer AIHero skill semantics and use OMX for support after acceptance criteria and verification are clear.
47 | - Keep OpenClaw, Codex, and OMX runtime state, credentials, tokens, transcripts, and logs out of the repository.
```

But search misses it:

```sh
[...]

Issue #506, opened 2026-05-27
Timeout on Zed


LeoMartinDev:
Hi,
The MCP server won't start properly on Zed IDE, it fails with a timeout error.

Here's the config I used:

```
    "codedb": {
      "command": "/home/leo/bin/codedb",
      "enabled": true,
      "args": [
        "mcp"
      ],
      "env": {}
    }
```

Issue #505, opened 2026-05-27
opencode error


eramax:
I tried to install it in my opencode by this config

```js
  "mcp": {
    "codedb": {
      "command": "/home/abb/bin/codedb",
      "args": [
        "mcp"
      ],
      "type": "local",
      "enabled": true
    }
  }
```
But i cannot find it in my mcp list in opencode.

and when i pushed opencode to use it i got 

> No MCP tools. So I can't directly use codedb's MCP server. But I can use the codedb CLI via bash.
Let me tell the user that codedb is configured as an MCP server in this project, and I can also use its CLI commands via bash.
It's configured as an MCP server in this project's opencode.json:6 — the binary lives at /home/abb/bin/codedb. I don't have direct MCP tool access, but I can invoke its CLI commands (like tree, outline, find, search, word, read, hot) via bash. Want me to run something with it

eramax:
This is how it worked for a ref.

```json
{
  "$schema": "https://opencode.ai/config.json",
  "mcp": {
    "codedb": {
      "type": "local",
      "command": [
        "npx",
        "-y",
        "codedeebee",
        "mcp"
      ],
      "enabled": true
    }
  }
}
```

PR #509, opened 2026-05-28
v0.2.5821 — 7 issue bundle (#501, #502, #503, #504, #505, #506, #507, #508)

### LINKED ISSUE #501: feat: add npm/npx distribution for codedb MCP usage
## Goal

Enable users to configure codedb in MCP clients with `npx` for zero-install setup:

```json
{
  "codedb": {
    "type": "local",
    "command": ["npx", "-y", "@justrach/codedb"],
    "args": ["mcp"],
    "enabled": true
  }
}
```

## Rationale

Many MCP clients (opencode, Claude Code, Cursor) document local servers through `npx`, especially for zero-install setup. codedb is currently a native Zig binary, which requires users to install it manually and hardcode an absolute binary path in their MCP config.

An npm package can be a **thin JS launcher** around the native binary. It does not require rewriting codedb in JavaScript.

## Requirements

- Publish `@justrach/codedb` with a `bin` entry named `codedb`
- JS wrapper must preserve:
  - `process.cwd()`
  - stdio (stdout reserved for JSON-RPC in MCP mode)
  - args
  - environment
- Wrapper must not write logs to stdout before MCP starts
- Provide platform binaries through either:
  - **optional platform packages** (preferred — `@justrach/codedb-darwin-arm64`, etc.)
  - or `postinstall` download from GitHub Releases
- Keep version aligned with codedb release version

## Suggested layout

```
npm/
  package.json
  bin/
    codedb.js        # thin launcher, execs native binary
  postinstall.js     # or rely on optionalDependencies
```

### Platform packages approach (recommended)

Top-level `package.json`:

```json
{
  "name": "@justrach/codedb",
  "version": "0.2.5820",
  "bin": { "codedb": "./bin/codedb.js" },
  "optionalDependencies": {
    "@justrach/codedb-darwin-arm64": "0.2.5820",
    "@justrach/codedb-darwin-x64": "0.2.5820",
    "@justrach/codedb-linux-x64": "0.2.5820",
    "@justrach/codedb-linux-arm64": "0.2.5820"
  }
}
```

### Launcher (`bin/codedb.js`)

```js
#!/usr/bin/env node
const { spawnSync } = require("node:child_process");
const path = require("node:path");

const exe = process.platform === "win32" ? "codedb.exe" : "codedb";
const bin = path.join(__dirname, "..", "vendor", `${process.platform}-${process.arch}`, exe);

const result = spawnSync(bin, process.argv.slice(2), {
  stdio: "inherit",
  cwd: process.cwd(),
  env: process.env
});
process.exit(result.status ?? 1);
```

## Acceptance criteria

1. This works from any directory:
   ```sh
   npx -y @justrach/codedb mcp
   ```

2. This works in opencode/Claude Code MCP config:
   ```json
   {
     "codedb": {
       "type": "local",
       "command": ["npx", "-y", "@justrach/codedb"],
       "args": ["mcp"],
       "enabled": true
     }
   }
   ```

3. `codedb_status` reports files from the current repo, not the npm cache directory

## Scope estimate

| Size | Conditions |
|------|-----------|
| **Small** | GitHub Actions already builds per-platform release binaries, maintainer has npm org access, package name available, no Windows initially |
| **Medium** | Need release asset naming convention, npm workspace setup, CI publish workflow, multi-platform optional packages |
| **Large** | Need full cross-platform binary build pipeline, signing/checksums, installer migration |

## Notes

- This is a packaging/distribution feature, not a code change to the Zig server
- Keep separate from MCP behavior fixes (e.g. #500)
- The hard part is reliable native binary distribution through npm, not MCP itself

justrach:
## Summary

Bundle of 7 fixes from the 2026-05-28 open-issue triage. All E2E verified, all 635/635 tests pass.

## Issues fixed

| # | Title | Fix |
|---|---|---|
| #501 | feat: add npm/npx distribution | Published `codedeebee` on npm; postinstall fetches matching native binary + sha256 verify; thin spawnSync launcher |
| #502 | mcp stuck in loading_snapshot + various sub-issues | parsePositional path-as-root + mcp --help bypass; `isValidMcpFlag` whitelist; `findGitRoot` walks up from cwd; deferred-loop gives up after 13s |
| #503 | `codedb mcp <path>` hangs | parsePositional special-cases the `mcp <path>` arg order |
| #504 | macOS Intel x64 segfault | Bisected via Rosetta: Zig 0.16's `!void` main runtime wrapper crashes on signed x86_64-macos. Fix: `pub fn main(...) void` + `mainTrampoline()`. Also adds `handleFastPath` for bare/`--version` invocations + `Out.exitWithFlush` so error messages survive early-exit paths |
| #505 | opencode "No MCP tools" | protocolVersion negotiation — echo client's version instead of hardcoding `2025-06-18` |
| #506 | Zed MCP timeout | Same negotiation fix |
| #507 | search misses content after rebuild | Outline-only files now go into `skip_trigram_files` so tier 3 substring-scans them |
| #508 | codedb_remote HTTP 530 UX | `appendRemoteErrorHint` distinguishes Cloudflare 1033 / 404 / 429 / 5xx with actionable hints (server-side outage at api.wiki.codes is unchanged — UX-only) |

## Distribution / installer (bonus)

- `npm/codedeebee` package shipped (closes #501)
- `install/install.sh`: hook-priority race fixed — when a competing legacy-tools hook is already registered for the same event/matcher, codedb's hook is inserted at index 0 instead of appended
- `README.md` refresh: "Or via npm/npx" install section, 16 → 21 MCP tools everywhere, dropped removed tools, added missing tools, removed obsolete v0.2.579 hotfix block, added `codedb read` to CLI table

## Test plan

- [x] `zig build test --summary all` → **635/635 pass** across all 8 test binaries
- [x] `codedb mcp /tmp/nonexistent` → "cannot resolve root", exit 1 (was: hang forever)
- [x] `codedb mcp --help` → usage to stdout, exit 0 (was: started MCP server)
- [x] `codedb mcp --bogus` → "unknown flag for mcp: --bogus", exit 1 (was: silently ignored)
- [x] `codedb` bare on arm64 + x86_64 via Rosetta → usage to stderr, exit 1 (was: SIGSEGV on x86_64)
- [x] `codedb --version` on arm64 + Rosetta → "codedb 0.2.5821", exit 0
- [x] MCP `initialize` handshake: client 2024-11-05 → server echoes 2024-11-05; 2025-03-26 → echoed; 2025-06-18 → echoed; 2099-01-01 → falls back to 2025-06-18
- [x] `test issue-507` passes (outline-only files now search-visible)
- [x] `test issue-508` passes (Cloudflare/404/429/5xx hints)
- [ ] User #504 (arrrrny) to confirm on native Intel Mac after release
- [ ] Notarize + upload binaries from local

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,30 @@
 # Changelog
 
 
+## 0.2.5821 - 2026-05-28
+
+Bundle of seven fixes from the open-issue triage on 2026-05-28.
+
+### MCP server fixes
+
+- **#502 + #503 — arg parser overhaul.** `codedb mcp <path>` no longer hangs forever in deferred mode (it now honors the path as root). `codedb mcp --help` prints usage instead of starting the server. Unknown post-`mcp` flags (e.g. `codedb mcp --snapshot`) are now rejected with a listed-valid-flags error. `codedb mcp` from a git-repo subdirectory walks up to the repo root. The deferred-scan path can no longer hang in `loading_snapshot` forever when the cwd isn't indexable — gives up after 13 s and unblocks `scan_done`.
+- **#505 + #506 — MCP protocol version negotiation.** The server previously hardcoded `protocolVersion: "2025-06-18"`, which older Zed and certain opencode versions rejected with a startup timeout / "No MCP tools". Now echoes the client's version when it's one we've verified against (`2024-11-05`, `2025-03-26`, `2025-06-18`); for newer-than-known clients we return our latest known version.
+- **#507 — search misses content after snapshot rebuild.** Files routed through `indexFileOutlineOnly` (snapshot load fallback, watcher incremental updates, WASM fast-path) were registered in `outlines` and `contents` but not in any search index. They were invisible to every search tier — including the tier-5 full-scan fallback, which short-circuited because the trigram index returned a non-null empty candidate set. Fixed by registering outline-only files in `skip_trigram_files` so tier 3 substring-scans them.
+- **#508 — actionable `codedb_remote` errors.** The remote tool now distinguishes Cloudflare 530 / 1033 origin-unreachable from 404 (repo not indexed), 429 (rate limited), and 5xx (upstream error) with retry / local-fallback hints. The server-side outage at `api.wiki.codes` is not fixed by this change; the UX is.
+
+### Startup / platform
+
+- **#504 — macOS Intel x64 segfault on bare `codedb`.** Bisected via Rosetta: Zig 0.16's runtime wrapper around `pub fn main(...) !void` crashes at startup on signed x86_64-macos binaries. The user saw `codedb` segfault before any output reached the terminal. Fix: `pub fn main(...) void` (infallible) + `mainTrampoline()` for the fallible work + a `handleFastPath` short-circuit for bare/`--version` invocations that writes via raw `std.c.write` and bypasses the worker-thread trampoline entirely. Also fixes a related "output silently lost on early exit" bug where `std.process.exit(_)` skipped the deferred `Out.flush()`; `Out.exitWithFlush` now handles the common usage / error-message exit paths.
+
+### Distribution
+
+- **#501 — npm/npx distribution.** Published [`codedeebee`](https://www.npmjs.com/package/codedeebee) as the npx-friendly sibling of `codedb`. `npx -y codedeebee mcp` does a one-shot install: thin Node launcher + `postinstall` that downloads the matching native binary from this GitHub release and SHA256-verifies against `checksums.sha256`. The bare `codedb` name is restricted on npm; the package is `codedeebee` but the CLI it installs is still called `codedb`.
+
+### Installer
+
+- **Hook-priority race.** `install/install.sh` now detects competing legacy-tools hooks (`block-legacy-tools.sh`, muonry, zigrep, zigread) and inserts codedb's hook at index 0 instead of appending. Re-runs reshuffle an already-registered codedb hook to the front if a competitor has appeared since the previous install.
+
+
 ## 0.2.5813 - 2026-05-12
 
 `0.2.5813` ships three structural improvements: a Tier 0 search-quality rewrite, a 4-6x faster regex matcher, and a bounded-memory content cache.
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@
 
 | What works today                                       | What's in progress                       |
 |--------------------------------------------------------|------------------------------------------|
-| 16 MCP tools for full codebase intelligence            | Deeper parser coverage and edge-case handling |
+| 21 MCP tools for full codebase intelligence            | Deeper parser coverage and edge-case handling |
 | Trigram v2: integer doc IDs, batch-accumulate, merge intersect | Incremental segment-based indexing |
 | 538x faster than ripgrep on pre-indexed queries        | WASM target for Cloudflare Workers       |
 | O(1) inverted word index for identifier lookup         | Multi-project support                    |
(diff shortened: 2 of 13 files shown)


yanncabral:
Thanks for the #504 fix in this PR. I tested the published `v0.2.5821` release asset on a native Intel macOS machine, and the issue still reproduces there.

Environment:

- macOS 26.2 (25C56)
- MacBookPro16,1
- `uname -m`: `x86_64`
- asset: `codedb-darwin-x86_64`
- sha256: `46bc218b55c0ba2aa404998c7d74d6ee3f2b8022b7dea23c9b2f3cf71b22213b`

Repro:

```sh
curl -fsSLO https://github.com/justrach/codedb/releases/download/v0.2.5821/codedb-darwin-x86_64
chmod +x codedb-darwin-x86_64
./codedb-darwin-x86_64 --help
echo $?
# 139
```

This is also the same binary installed by `npx -y codedeebee mcp`, so MCP clients launching through `npx` still fail with `Connection closed` because the native server exits immediately.

Crash report summary:

```text
EXC_BAD_ACCESS / SIGSEGV
KERN_INVALID_ADDRESS at 0x0000000000000000
faultingThread: 0
```

One useful data point: building current `main` locally on the same machine works:

```sh
git clone https://github.com/justrach/codedb.git
cd codedb
zig build -Doptimize=ReleaseSafe
./zig-out/bin/codedb --help
# exit 0
```

So the source-level fix appears to be present on `main`, but the published `v0.2.5821` `codedb-darwin-x86_64` release asset still crashes on native Intel macOS.

I also added this reproduction info to #504: https://github.com/justrach/codedb/issues/504#issuecomment-4567540854

yanncabral:
Follow-up with a smaller repro and a likely direction for the fix.

I was able to reduce this outside of `codedb`: on native Intel macOS 26.2, a minimal Zig 0.16 `x86_64-macos` executable works unsigned, but segfaults after `codesign`.

Environment:

```text
macOS 26.2 (25C56)
MacBookPro16,1
uname -m: x86_64
zig: 0.16.0 (/usr/local/Cellar/zig/0.16.0_1/bin/zig)
```

Minimal repro:

```zig
const std = @import("std");

pub fn main() void {
    const msg = "hello\n";
    _ = std.c.write(1, msg.ptr, msg.len);
}
```

Commands:

```sh
zig build-exe hello.zig -O ReleaseFast -target x86_64-macos -lc -femit-bin=hello
./hello
echo $?

codesign -f -s - hello
./hello
echo $?
```

Observed:

```text
hello
0
hello
Segmentation fault: 11
139
```

So `main` does run and prints `hello`, but the signed Zig binary crashes during/after teardown. The crash report summary is:

```text
EXC_BAD_ACCESS / SIGSEGV
KERN_INVALID_ADDRESS at 0x0000000000000001
```

I also tried these variants, and all signed Zig binaries still crashed:

```sh
-fllvm
-flibllvm
-fno-lld
-funwind-tables
-fno-unwind-tables
-target x86_64-macos.10.15
-target x86_64-macos.11.0
-target x86_64-macos.13.0
-target x86_64-macos.26.2
```

A C control binary signed with the same `codesign -f -s -` works on the same machine:

```c
#include <stdio.h>
int main() { puts("hello-c"); return 0; }
```

```sh
cc hello.c -o hello-c
codesign -f -s - hello-c
./hello-c
# hello-c
# exit 0
```

This suggests the `codedb` crash is probably not project-specific and not just the `!void` entrypoint issue. It looks like a Zig 0.16 / Mach-O / x86_64 macOS 26 issue that is triggered once the binary is signed.

I also built `codedb` with a small release-pipeline workaround: skip codesigning only for the `x86_64-macos` slice. That produced an unsigned `codedb` binary that works on this machine:

```sh
zig build -Doptimize=ReleaseFast -Dtarget=x86_64-macos -Dcodesign-identity=-

codesign -dv --verbose=2 zig-out/bin/codedb
# code object is not signed at all

./zig-out/bin/codedb --help
# exit 0

./zig-out/bin/codedb --version
# exit 0
# codedb 0.2.5821
```

MCP smoke test also worked with that unsigned binary:

```text
initialize -> result.protocolVersion = 2024-11-05
serverInfo.name = codedb
serverInfo.version = 0.2.5821
scan = ready
files = 377
```

The workaround patch I tested was basically:

- `build.zig`: make `codesign-identity` opt-in instead of defaulting to `-`, and skip signing when `target.result.cpu.arch == .x86_64` on macOS.
- `.github/workflows/release-binaries.yml`: do not pass `-Dcodesign-identity` for the `x86_64-macos` matrix entry.

This is not ideal for macOS distribution, but it may be a practical short-term workaround for the npm/MCP path until the Zig/Mach-O signing issue is understood. I tried to file this minimal repro upstream at `ziglang/zig`, but interactions there are currently restricted to collaborators, so GitHub blocked both issue creation and commenting on the related Tahoe/x86_64 issue.

PR #517, opened 2026-05-31
feat: edit syntax-health checks + anchored str_replace + opt-in linter bridge

justrach:
## What

Three layers that make `codedb_edit` safer, plus an opt-in real-linter bridge. Everything is advisory and **off the edit hot path**; the external linters are **off unless you opt in** at `codedb update`, so default behavior + latency are unchanged.

**1. Post-edit syntax health (in-process, always on, microseconds)**
After every edit, codedb scans the result and warns if the edit broke the file:
- unbalanced/duplicated delimiter (e.g. a stray `)` from regenerating a multi-line import block),
- an import name removed while still referenced (a `NameError` waiting to happen).

**2. Anchored `str_replace`**
`codedb_edit` gains `op=str_replace` (`old_string`/`new_string`) — an exact, unique byte splice that can't mis-target surrounding lines the way a range replace can. Advertised as the preferred edit.

**3. External-linter bridge (opt-in, off the hot path)**
When enabled, codedb runs a real linter on the edited file on a **detached thread** (never blocking the edit) and serves results via a new `codedb_diagnostics` tool (+ piggyback on the next edit). Registry: ruff (Python), biome (JS/TS/CSS), `zig ast-check`, gofmt (Go), `ruby -c`, `php -l`, cppcheck (C/C++), shellcheck (shell), ktlint (Kotlin), swiftlint (Swift). Opt-in is asked once at `codedb update` (TTY only; the MCP server never prompts/installs), prefers **nanobrew** when present, and falls back to uv/npm/brew. A missing/failed linter is disabled for the session, so it silently falls back to the in-process checks.

## Why

Functional validation of the deep-SWE benchmark found the codedb-driven agent shipped patches that don't even import (httpx: a duplicate `)`; narwhals: a dropped-but-used import producing a `NameError`). Layer 1 catches exactly that class at edit time instead of at runtime.

## Safety / latency

- Off by default; when off, the edit path does zero extra work (the Tier-1 block is gated behind `enabled`).
- When on, the linter runs on a detached thread after the edit response is built — no added edit latency. A bounded diagnostics cache drains in-flight workers before teardown (no use-after-free).
- The always-on in-process check is a single ~microsecond byte scan.

## Testing

`zig build test` green. Live MCP smoke verified end-to-end on **macOS** (ruff F401/F821, `zig ast-check`, shellcheck SC2154/SC2086) and **Linux** (cross-compiled static `aarch64-linux-musl`, run under Apple `container`), including the missing-tool fallback.

## Needs reconciliation with current `main`

This branch was cut from `main` **before 0.2.5822**. Since then `main` rewrote `applyEdit` (path resolution via `explorer.root_dir`, two fast-path replace blocks, a new `EditResult.changed` field) and grew `mcp.zig` substantially. Rebasing onto current `main` produces a **2-hunk conflict in `src/edit.zig` only** — `mcp.zig`/`update.zig` auto-merge. Reconciliation plan: re-apply `str_replace` + the health check onto `main`'s `applyEdit`, route all write paths through one `finalizeEdit`, and keep `main`'s `.changed` / fast-path optimizations. Happy to push that rebase on request.

## Commits

13 focused commits: P0a delimiter scan, P0b dropped-import scan, P2 anchored str_replace, linter registry/preference/exec/prompt, DiagnosticsCache, daemon wiring + `codedb_diagnostics`, nanobrew install option, and shellcheck/cppcheck/ktlint/swiftlint + ANSI-strip.


justrach:
Closing in favor of release/0.2.5824, which already contains these commits; the linter-bridge/syntax-health work will reach main when the release merges. Branch deleted during stale-branch cleanup.

Issue #518, opened 2026-06-01
codedb 0.2.5823: a few correctness/UX findings (non-ASCII outline, codedb_find false hits, kind labels, search cap, snapshot staleness)


ahndohun:
## codedb v0.2.5823 — a few correctness/UX findings from a controlled audit

Hi — Pro user here, big fan of codedb (symbol recall was effectively 100% to the exact file:line in our tests, `codedb_word` is exhaustive and fast, and `codedb_outline` matched Python `ast` exactly on real files). While auditing we hit a few smaller things and would like to know if they're bugs, intended, or our misuse. Environment: codedb 0.2.5823, macOS arm64; ground truth computed with Python `ast`/`re`.

### 1. `codedb_outline` / `codedb_symbol` return nothing for non-ASCII (e.g. Korean) identifiers
```sh
printf 'def \xed\x95\x9c():\n    return 1\n' > /tmp/uni.py    # "def 한():"
# index a folder containing it, then:
# codedb_outline uni.py        -> header only, 0 symbols
# codedb_symbol  한            -> "no results"
# codedb_search/word for the bytes DOES find it
```
Python `ast` parses the function fine (valid Python 3 identifier). Is non-ASCII identifier support intended for the structural layer?

### 2. `codedb_find` returns confident hits for queries that match no filename (no score floor)
```
codedb_find "zzznosuchfilexyz"  -> notrail.py (32.79), oracle.json (31.36), unicode.py (25.30) ...
codedb_find "Widget"            -> empty.py (28.29), crlf.txt (24.96) ...   (no file named Widget)
```
The fuzzy subsequence matcher never returns empty, so a non-match looks like a ranked result. A score floor (or an explicit "no match") would help callers distinguish.

### 3. Kind labeling: Python `class` shown as `struct_def` / uniform `fn` header
`codedb_symbol Widget` on a Python `class Widget:` reports kind `struct_def`, and the status header labels every symbol `fn` regardless of type. `ast` says `class`. Cosmetic, but it confuses class-vs-function navigation in a Python codebase.

### 4. `codedb_search` caps at 50 results — is `codedb_word` the intended exhaustive tool?
For a token occurring 180x, `codedb_search` returns at most 50 (default 20), while `codedb_word` returned all 180+ uncapped. We've adopted `codedb_word` for exhaustive single-identifier lookups — just confirming that's the intended pattern, and whether the `codedb_search` cap is configurable.

[...]

PR #526, opened 2026-06-03
find: faster + smarter fuzzy file ranking (prefilter + SIMD + symbol fast-path)

justrach:
## Summary

`codedb_find` (fuzzy filename search) ran a per-file Smith-Waterman DP over every file in the tree. On large repos the long, symbol-like queries that don't hit the exact-filename fast path were slow (openclaw, 13.6k files: ~250ms Debug / ~41ms ReleaseFast for `getTokenProvider`). Real filename queries (`auth`, `config`, `index`) were already <0.5ms — only the degenerate "used find as a symbol lookup" case was slow.

This branch speeds up that case **with zero change to which files come back or their order**, verified on both a unit matrix and end-to-end on openclaw (recall@10 = recall@30 = 1.00, exact result order preserved vs the release baseline across 16 representative queries).

## Two commits

**1. Sound presence prefilter + tightened scalar DP** (`55df894`)
- Issue #518 made a ≥60% in-order LCS a hard requirement. Unordered character presence is an upper bound on that LCS, so a path missing >40% of the query's chars can be rejected in O(path) without the DP — it would fail the LCS gate anyway, so no result changes.
- Hoisted per-row lowercasing, precomputed path lowercase + word-boundary flags, swapped the DP rows by pointer (no `@memcpy`), folded the matched-row scan into the main loop.

**2. SIMD-across-files Smith-Waterman** (`0e954b6`)
- Refactored the scalar scorer into `fuzzyDP` (hot loop) + `fuzzyFinalize` (floor/LCS/bonus tail).
- Added `fuzzyScoreBatch`: the same DP vectorized so each of 8 SIMD lanes scores a **different file** against the same query (inter-sequence parallelism). Shorter paths are length-masked per column so padding never affects `best_score`/`matched_chars`; the shared `fuzzyFinalize` tail keeps post-processing identical.
- Single-part queries (the common case) batch presence-survivors 8 at a time; multi-part falls back to scalar.
- All DP values are sums of exactly-representable small integers → the SIMD result is **bit-identical** to scalar.

## Measured (ReleaseFast, openclaw 13.6k files, interleaved vs release baseline)

| query | baseline | this PR | speedup | recall@10 | order |
|---|---|---|---|---|---|
| getTokenProvider | 41 ms | 23 ms | 1.80× | 1.00 | exact |
| handleRequest | 33 ms | 19 ms | 1.78× | 1.00 | exact |
| embedded | 22 ms | 9 ms | 2.39× | 1.00 | exact |
| subscribe | 25 ms | 14 ms | 1.82× | 1.00 | exact |
| auth / config / index | <0.5 ms | <0.5 ms | — | 1.00 | exact |

8 lanes beat 16 on Apple-Silicon NEON (register pressure). The remaining gap from a theoretical 4× is the per-batch transpose plus the DP's serial per-column dependency.

## Tests
New `test "fuzzy SIMD batch scorer matches scalar fuzzyScore exactly"` compares `fuzzyScoreBatch` to `fuzzyScore` over a query×path matrix (incl. a full 8-lane mixed-length batch). Full suite green.

## Investigated and rejected
A cheap candidate cap (greedy proxy → DP top-N) reached ~10× but dropped recall@10 to 0.85 (0.00 on `getTokenProvider`) — the greedy proxy can't model Smith-Waterman's mismatch tolerance. Dropped in favor of the retrieval-identical SIMD path.

### DIFF
--- a/src/explore.zig
+++ b/src/explore.zig
@@ -3122,26 +3122,11 @@ pub const Explorer = struct {
         var iter = self.outlines.keyIterator();
         while (iter.next()) |key_ptr| {
             const path = key_ptr.*;
-
-            // Extension filter
             if (ext_filter) |ext| {
                 if (!std.mem.endsWith(u8, path, ext)) continue;
             }
-
-            // Multi-part scoring: all parts must match, scores sum
-            var total_score: f32 = 0;
-            var all_matched = true;
-            for (parts.items) |part| {
-                if (fuzzyScore(part, path)) |s| {
-                    total_score += s;
-                } else {
-                    all_matched = false;
-                    break;
-                }
-            }
-
-            if (all_matched and total_score > 0) {
-                try matches.append(allocator, .{ .path = path, .score = total_score });
+            if (scoreAllParts(parts.items, path)) |s| {
+                try matches.append(allocator, .{ .path = path, .score = s });
             }
         }
 
@@ -5862,10 +5847,39 @@ fn lcsLenIgnoreCase(query: []const u8, path: []const u8) usize {
     return row[pn];
 }
 
+// Score `path` against every space-separated query part (all must match, scores
+// sum) — the exhaustive per-file scoring shared by both fuzzyFindFiles passes.
+fn scoreAllParts(parts: []const []const u8, path: []const u8) ?f32 {
+    var total: f32 = 0;
+    for (parts) |part| {
+        if (fuzzyScore(part, path)) |s| {
+            total += s;
+        } else return null;
+    }
+    if (total > 0) return total;
+    return null;
+}
 pub fn fuzzyScore(query: []const u8, path: []const u8) ?f32 {
     if (query.len == 0 or path.len == 0) return null;
     if (query.len > 128 or path.len > 512) return null;
 
+    // Cheap presence pre-filter. The score floor below (issue #518) makes a
+    // >=60% in-order LCS between query and path a hard requirement. Unordered
+    // character presence is an upper bound on that LCS, so if fewer than 60% of
+    // the query's characters occur anywhere in the path, the Smith-Waterman DP
+    // cannot clear the floor and we can reject in O(path) instead of running the
+    // O(query*path) DP. This skips the DP for the vast majority of non-matching
+    // files (codedb_find on a 670-file repo: ~265ms -> a few ms) and changes no
+    // result: every file rejected here would also fail the LCS check at the end.
+    {
+        var present = [_]bool{false} ** 256;
+        for (path) |c| present[toLowerByte(c)] = true;
+        var hits: usize = 0;
+        for (query) |c| {
+            if (present[toLowerByte(c)]) hits += 1;
+        }
+        if (hits * 5 < query.len * 3) return null;
+    }
     const MATCH_SCORE: f32 = 16.0;
     const MISMATCH_PENALTY: f32 = -8.0;
     const GAP_OPEN: f32 = -3.0;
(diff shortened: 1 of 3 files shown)


justrach:
## Added: symbol fast-path (3rd commit, 41f79d4)

The slow `codedb_find` queries were symbol-like identifiers (`getTokenProvider`, `handleRequest`) typed into a *fuzzy filename* search — they match no filename, so they paid the full scan **and** returned useless filename guesses rather than the definition the caller wanted.

`handleFind` now, after the exact-filename check and before the fuzzy scan, detects a **compound identifier** (camelCase/PascalCase or snake_case; never ALL-CAPS, dotted, slashed, spaced, or <4 chars) and resolves it against the symbol index, returning the **definition site(s)**. Non-symbol identifiers fall through to the (SIMD-accelerated) fuzzy scan, so real filename searches are unaffected.

Uses the **complete** `findAllSymbols` lookup (incl. the outline safety scan), not `symbol_index` alone — `symbol_index` is incomplete for some languages (it indexes ~no TypeScript functions; a `symbol_index`-only version scored **0/40** real openclaw symbols). Definition kinds rank above import/comment usages.

**Measured (ReleaseFast, openclaw 13.6k files):** 40/40 real symbols resolve via the fast-path at **~1.3 ms** (returning the def location) vs **~29 ms** on the fuzzy path — **~22× faster and more useful**. Filename queries (`auth`, `config`) keep the <0.5 ms exact/fuzzy path.

New test: classifier truth table + `renderSymbolDefsFast` resolves a def and returns false (no output) on a miss. 700+ tests green.

### Net effect of this PR on `codedb_find`
- **Real symbol typed into find** (the main slow/misused case): ~29 ms fuzzy guess → **~1.3 ms exact definition** (~22×).
- **Compound query that isn't a symbol:** fuzzy scan, now SIMD-accelerated (~1.8× on the heavy ones).
- **Filename query:** unchanged, <0.5 ms.

justrach:
## What makes it faster (and what's unchanged)

Four mechanisms, in the order `handleFind` applies them:

1. **Symbol fast-path (the big structural win).** A compound-identifier query (camelCase/snake_case) is resolved against the symbol table and returns the **definition site** — it *skips the per-file fuzzy scan entirely*. This is an algorithm change (O(files) Smith-Waterman → O(symbol lookup)), not a constant-factor tweak: ~30–50× for those queries, and a *better* answer (the def, not filename guesses). Uses the complete `findAllSymbols` (incl. the outline safety scan) because `symbol_index` alone is incomplete for some languages (0/40 real TS symbols).
2. **Sound presence pre-filter.** Issue #518 makes a ≥60% in-order LCS a hard requirement; unordered char-presence is an upper bound on that LCS, so files missing >40% of the query's chars are rejected in O(path) — skipping the O(query×path) DP for files that would fail the gate anyway. **Retrieval-identical.**
3. **SIMD-across-files Smith-Waterman.** The fuzzy fallback scores 8 files per vector pass (one file per lane). All DP values are sums of exactly-representable small ints → **bit-identical** to scalar. ~2× on a cool machine.
4. **Tightened scalar DP.** Hoisted the per-row lowercased query char, precomputed path-lowercase + word-boundary flags, pointer-swapped the DP rows (no `@memcpy`), folded the matched-row scan.

## Verification

- **1:1 with everything else.** Data output of all 10 non-find tools (search, callers, symbol, word, deps, tree, outline, glob, context) is **byte-identical** baseline vs this branch — find is the only behavior change, and its fuzzy path is retrieval-identical (recall@10/@30 = 1.00, exact order).
- **RSS flat.** 157 MB vs baseline 158 MB after 200 mixed find queries — the SIMD lane arrays + candidate list are transient per-query (freed each call).
- **Tests:** 700+ green.
- **Both surfaces:** lives in `handleFind`, shared by MCP `codedb_find` and CLI `file` (warm daemon → `runCliTool` → `handleFind`). Measured CLI symbol query: ~40–60 ms → <4 ms.
- **Measurement caveat:** the per-query fuzzy speedup is sensitive to this box's documented macOS E-core/thermal variance; using `config`/`auth` (untouched code) as a control, every fuzzy query's new/base ratio sits *at or below* the control penalty, confirming the SIMD path is not a regression.

PR #535, opened 2026-06-05
feat(remote): add local fallback when api.wiki.codes is unreachable

### LINKED ISSUE #534: codedb_remote: add local fallback when api.wiki.codes is unreachable
## Problem

`codedb_remote` depends on `api.wiki.codes`, which is frequently unreachable (Cloudflare 530 errors, rate limiting, timeouts). This makes the remote tool unreliable in practice, while the local tools (tree, outline, search, symbol, etc.) work great.

## Suggested approach

Add a fallback chain when `api.wiki.codes` fails:

1. **Try remote first** (current behavior)
2. **Check local cache**: `~/.codedb/remote-cache/<repo-hash>/`
   - If exists and fresh (TTL), use the cached local index
   - If missing or stale, `git clone --depth=1` + `codedb index`
3. **Query local index** with the same action/params

### Cache management
- **TTL**: auto-delete repos not accessed in ~7-30 days
- **LRU**: cap at N repos or max disk size (e.g. 2GB)
- **Cleanup**: run lazily on query or as a background task
- **Manual**: `codedb remote --clean-cache`

### Benefits
- Self-healing when the upstream is down
- Fast after first clone (cached locally)
- Shallow clones keep disk usage reasonable
- Works offline after initial clone

### Trade-offs
- First query is slow (clone + index, 10-60s for large repos)
- Auth needed for private repos (GitHub token)

This would make `codedb_remote` genuinely robust instead of a single point of failure.

idea404:
Closes #534

When api.wiki.codes is unreachable, automatically fall back to a local shallow clone + index for tree, outline, search, read, symbol, and deps actions.

**Changes:**
- New `src/remote_cache.zig` module for cache management (`~/.codedb/remote-cache/`)
- Modified `handleRemote` to try remote first, then local fallback on failure
- Added `clean_cache` action to purge cached repos
- Updated tool schema to document fallback behavior

**Fallback chain:**
1. Try `api.wiki.codes` (existing behavior)
2. On failure, check if action is locally serviceable
3. Ensure repo is cached locally (`git clone --depth=1` + `codedb snapshot`)
4. Load cached snapshot and dispatch to local handlers

**Verification:**
- `zig build` — compiles clean
- `zig build test` — 709/710 pass (1 pre-existing flaky perf test)
- E2E MCP test — 20/20 pass

### DIFF
--- a/src/lib.zig
+++ b/src/lib.zig
@@ -60,3 +60,4 @@ pub const watcher = @import("watcher.zig");
 pub const mcp = @import("mcp.zig");
 pub const snapshot = @import("snapshot.zig");
 pub const snapshot_json = @import("snapshot_json.zig");
+pub const remote_cache = @import("remote_cache.zig");
--- a/src/mcp.zig
+++ b/src/mcp.zig
@@ -3685,22 +3685,58 @@ fn handleRemoteFallback(
         loadProjectWordIndexFromDiskIfPresent(io, ctx.explorer, cache_dir, alloc);
     }
 
+    var translated_args: std.json.ObjectMap = .empty;
+    defer translated_args.deinit(alloc);
+    translateRemoteArgs(alloc, args, action, &translated_args);
+
     if (std.mem.eql(u8, action, "tree")) {
         handleTree(alloc, out, ctx.explorer);
     } else if (std.mem.eql(u8, action, "outline")) {
-        handleOutline(alloc, args, out, ctx.explorer);
+        handleOutline(alloc, &translated_args, out, ctx.explorer);
     } else if (std.mem.eql(u8, action, "search")) {
-        handleSearch(alloc, args, out, ctx.explorer);
+        handleSearch(alloc, &translated_args, out, ctx.explorer);
     } else if (std.mem.eql(u8, action, "read")) {
-        handleRead(io, alloc, args, out, ctx.explorer);
+        handleRead(io, alloc, &translated_args, out, ctx.explorer);
     } else if (std.mem.eql(u8, action, "symbol")) {
-        handleSymbol(alloc, args, out, ctx.explorer);
+        handleSymbol(alloc, &translated_args, out, ctx.explorer);
     } else if (std.mem.eql(u8, action, "deps")) {
-        handleDeps(alloc, args, out, ctx.explorer);
+        handleDeps(alloc, &translated_args, out, ctx.explorer);
     } else {
         return false;
     }
     return true;
+}
+
+pub fn translateRemoteArgs(alloc: std.mem.Allocator, args: *const std.json.ObjectMap, action: []const u8, out: *std.json.ObjectMap) void {
+    if (std.mem.eql(u8, action, "outline")) {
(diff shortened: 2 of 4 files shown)


justrach:
Hi @idea404 — thank you for tackling this! 🙏 The api.wiki.codes outages (#508/#534) are a real pain, and a local clone-and-index fallback is a genuinely good answer.

**1. Could you retarget the base `main` → `release/0.2.5825`?** (Edit → base branch.) We integrate on the release branch — we've added a CI hint that'll nudge this automatically going forward.

**2. Verified:** it compiles and the full suite stays green, no regressions. The security-sensitive parts are well done — `git clone` via an argv array (no shell injection), URL hardcoded to `github.com` (no SSRF), and `wikiSlugForRepo` rejects `..`/`//` so the cache path can't traverse out (the failure-path `deleteTree` is safe).

**A few things before we can merge:**
1. **No test for the new behavior** — the clone-fallback path is unexercised; our repo rule is every PR ships a test (one that forces the upstream unreachable and asserts the local fallback serves results would do it).
2. **`MAX_CACHED_REPOS` + the TTL helpers (`cleanStaleEntries`/`isCacheFresh`) are never called** — `ensureCached` gates only on the snapshot *existing*, so the cache grows unbounded and never expires.
3. **No clone timeout / size cap** — a huge or hanging repo would block the handler and fill the disk when the fallback fires.
4. **Concurrent fallback for the same repo** can have one clone `deleteTree` another's dir — a lock or clone-to-temp-then-rename fixes it.

All very doable and we're happy to help. If we don't hear back in ~1 day, we may pick these up in a follow-up building on your work, **with you credited (author/co-author + release notes)**. Thanks again! 🙏

idea404:
Thanks for the thorough review! All four points addressed in 774bd1e:

1. **Retargeted** to `release/0.2.5825`.

2. **Tests added** — `isCacheFresh` TTL enforcement and clone-to-temp-then-rename atomicity in `test_mcp.zig`.

3. **TTL + LRU wired up** — `ensureCached` now checks freshness (not just existence), runs `cleanStaleEntries` before cloning, and `evictIfNeeded` enforces `MAX_CACHED_REPOS` via LRU eviction.

4. **Clone timeout + size cap** — `git -c http.lowSpeedLimit=1000 -c http.lowSpeedTime=30` aborts stalled clones; `--single-branch --no-tags` keeps size down.

5. **Race fix** — clone to `<dir>.tmp.<timestamp>`, then atomic `rename` to final path. Failed clones clean up the temp dir only.

idea404:
Also addressed the Codex bot suggestions — `translateRemoteArgs` now maps `query→path` (outline), `query→name` (symbol), and `lines "10-60"→line_start/line_end` (read) before dispatching to local handlers.

idea404:
@justrach quick bump

justrach:
Closing: this content already shipped — merged into main as 13366d0 ("merge PR #535: codedb_remote local-clone fallback for #508") and the head branch has zero diff against release/0.2.5833-mcp2026. The PR was left open because the merge was done manually.

PR #542, opened 2026-06-07
Fix multi-line TS/JS imports getting dropped from the dep graph

### LINKED ISSUE #540: Multi-line TS/JS imports are missing from the dependency graph
The TS/JS parser only records an import when `from "..."` is on the same line as `import`. When an import spans multiple lines, the `from "..."` ends up alone on the closing `}` line, and that line never gets parsed as an import. The dependency is silently dropped.

This hits almost every formatted codebase, since Prettier and Biome both wrap imports like this:

```ts
import {
  a,
  b,
} from "./thing.ts";
```

`codedb deps file.ts --depends-on` shows nothing for `./thing.ts`. Single-line imports work. Multi-line ones disappear.

Cause: `parseTsLine` only looks for the string literal on lines that contain `import `. The `} from "..."` continuation line doesn't, so it falls through. The Go parser already tracks multi-line import blocks. TS has no equivalent.

nsxdavid:
Fixes #540.

`parseTsLine` only grabbed the import path off lines that contained `import `. Multi-line imports put `from "..."` on the closing `}` line, so the dependency got dropped. This adds a continuation branch that pulls the path off `} from "..."` lines, plus `export ... from` re-exports which were dropped the same way.

The branch is gated on those concrete shapes (`} from ` and `export ... from`) rather than any line containing ` from "`, so strings and comments like `select * from "users"` don't get mistaken for dependencies.

Red to green: the new test indexes a multi-line import, a re-export, and a comment containing `from "config.json"`. It checks the real imports land in `outline.imports` and the comment does not. Fails before the change (multi-line path missing), passes after. The rest of the suite stays green.

Files: `src/explore.zig` (parseTsLine), `src/test_parser.zig`.

One function plus a test, no generated artifacts, branch is off current main. Read CONTRIBUTING and followed it.

### DIFF
--- a/src/explore.zig
+++ b/src/explore.zig
@@ -3610,6 +3610,22 @@ pub const Explorer = struct {
             if (extractStringLiteral(line)) |path| {
                 try appendImportPath(a, outline, path);
             }
+        } else if (startsWith(line, "} from ") or
+            (startsWith(line, "export ") and containsAny(line, &.{ " from \"", " from '" })))
+        {
+            // Closing line of a multi-line import/export, e.g.
+            //   import {
+            //     a, b,
+            //   } from "../mod.ts";
+            // or a re-export like `export * from "./x.ts"`. The `from "..."`
+            // clause lands on a line that does not contain "import ", so without
+            // this branch the path is dropped (only single-line imports were
+            // captured). Gated on these concrete shapes so arbitrary `from "..."`
+            // inside strings/comments (e.g. SQL `select * from "users"`) is not
+            // mistaken for a dependency.
+            if (extractStringLiteral(line)) |path| {
+                try appendImportPath(a, outline, path);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


justrach:
Hi @nsxdavid — thank you for this! 🙏 Clean fix, and guarding against `from "..."` inside comments/strings was exactly the right call.

Quick logistics note: we land changes on the active **`release/*`** branch rather than `main` (just added a CI hint so this is clearer going forward — sorry for the friction). Rather than make you retarget, I've cherry-picked your commit straight onto **`release/0.2.5825`** as `5cd1e6d`, **with your authorship preserved**. Verified there: clean apply, builds on Zig 0.16, your `issue-1` test passes, full suite green (725/725).

You're credited as the commit author and will be listed in the **0.2.5825 release notes**. Closing this since it's landed there. Thanks again for the contribution! 🙌

nsxdavid:
Cool... for the future: should I make PRs to the active release/* branch instead?  Not sure if I missed something in the CONTRIBUTING docs.

Issue #553, opened 2026-06-07
codedb . status leaks a ~20GB resident process that never exits (warmup -> OOM/freeze)


lekt9:
## Summary

The `SessionStart` warmup pattern `codedb . status &` leaks a **~20 GB resident process that never exits**. Each invocation spawns another one, so on a machine where it runs repeatedly (e.g. a Claude Code `SessionStart` hook) the orphans accumulate until RAM is exhausted and macOS freezes the UI and force-restarts.

## Environment

- codedb **0.2.5824**
- macOS **26.0.1**, Apple Silicon, 64 GB unified memory
- Invoked from a Claude Code hook: `codedb . status >/dev/null 2>&1 &` on `SessionStart`

## What happens

- `codedb . status` appears to load the **full index into memory (~20 GB RSS; ~2.8 GB private/dirty + ~18 GB mmap'd index)** and then **does not exit** — the process stays resident for hours.
- When backgrounded (`&`) from a short-lived parent, it is reparented to `launchd` (`ppid=1`) and lives forever.
- Repeated invocations stack: I observed **3 simultaneous orphaned `codedb . status` processes at ~20 GB RSS each**, all `ppid=1`, plus the live `codedb mcp`.

Observed via `ps`:

```
pid=7591 ppid=1 up=07:04 rss=20.2GB  codedb . status
pid=9377 ppid=1 up=03:50 rss=20.7GB  codedb . status
pid=7585 ppid=7574       rss=20.1GB  codedb mcp     (live, expected)
```

## Impact

On a 64 GB machine this exhausts RAM. With swap effectively at 0, macOS responds with **jetsam kills** and **WindowServer `userspace_watchdog_timeout`** hangs — i.e. the screen freezes and the only recovery is a hard restart. Confirmed from `/Library/Logs/DiagnosticReports/`:

```
JetsamEvent-2026-06-07-170555.ips   (largestProcess = python3.12)
WindowServer_..._userspace_watchdog_timeout.spin   (x2)
```

## Reproduce

```sh
for i in 1 2 3; do codedb . status >/dev/null 2>&1 & done
sleep 5
ps -axo pid,ppid,rss,command | grep 'codedb . status'   # 3 procs, ~20 GB RSS each, none exit
```

## Expected

- `codedb . status` should be a **cheap, fast-exiting** status query — it should not need to load and hold the entire ~20 GB index resident, and it should **return promptly and exit**.
[...]

PR #556, opened 2026-06-08
fix(#553): cheap metadata-only CLI status — stop materializing the full index

justrach:
## Problem (#553, reported by @lekt9)

`codedb [root] status` is grouped with the heavy query commands, so it **materializes the full index before reading three cheap counts**:
- with a snapshot → mmaps the whole index (~18GB on the reporter's repo) + builds ~2.8GB in-memory;
- with no/stale snapshot → full `initialScan` + `buildCallCentrality` + a multi-GB snapshot rewrite (the multi-hour path);
- and because `status` is in `cliIsQueryCmd`, it **also auto-spawns a warm cli-daemon** — a second resident copy.

Run backgrounded from a Claude Code `SessionStart` hook (`codedb . status >/dev/null 2>&1 &`), each call reparents to launchd (ppid=1) and lingers. They stack across sessions until RAM is exhausted → jetsam kills + WindowServer `userspace_watchdog_timeout` freeze (hard restart). The reporter saw 3× ~20GB `codedb . status` orphans on a 64GB machine.

## Fix

An early `status` branch **before** the load/daemon path that reports from on-disk metadata only and exits immediately:
- `readStatusMeta` → `TrigramIndex.readDiskHeader` (header-only read of `trigram.postings`: `file_count` + `git_head`). No Explorer, no snapshot mmap, no re-index, **no daemon spawn**.
- No persisted index → prints **"not indexed — run `codedb <root> index`"** instead of triggering a full scan.

The `codedb_status` **MCP** tool is unaffected — it runs inside the already-loaded server, where the counts are free.

### Before → after
Old `status` loaded the full index (seconds-to-hours + GBs resident). New `status`, measured on this repo (670 files):
```
✓ codedb 0.2.5825  ⚡ 72.0µs
  root      .
  files     670 indexed
  head      1047ab9a0b2c
  data      ~/.codedb/projects/<hash>
```
72µs, exits immediately, no index load.

## Test

`test "issue-553: status reads file_count from disk header without loading the index"` (test_index.zig): asserts the cheap path reports `file_count`/`git_head` from the header for an indexed dir and `indexed=false` for an empty one. It references the new `readStatusMeta`, so it **cannot compile/pass without this fix**.

`zig build test` → **23/23 steps, 727/727 tests passed**.

## Notes
- Minimal/surgical: the rich status (outline count, trigram type, RSS) still renders inside the MCP server via `codedb_status`; only the cold CLI path changed.
- Reported on 0.2.5824; confirmed still present on 0.2.5825.

### DIFF
--- a/src/index.zig
+++ b/src/index.zig
@@ -1913,6 +1913,26 @@ pub const TrigramIndex = struct {
     }
 };
 
+/// On-disk `status` metadata, readable WITHOUT loading the full index into
+/// memory. #553: CLI `status` used to materialize the whole index (snapshot
+/// mmap, or a full re-index + multi-GB snapshot rewrite when none existed)
+/// before reading a few counts, so a backgrounded `codedb . status &` from a
+/// SessionStart hook leaked a multi-GB resident orphan that stacked to OOM.
+pub const StatusMeta = struct {
+    indexed: bool,
+    file_count: u32,
+    git_head: ?[40]u8,
+};
+
+/// Read `status` metadata from the persisted trigram header alone — no Explorer,
+/// no snapshot load, no re-index. Returns indexed=false when the project has no
+/// persisted index so the caller reports "not indexed" instead of re-indexing.
+pub fn readStatusMeta(io: std.Io, data_dir: []const u8, allocator: std.mem.Allocator) StatusMeta {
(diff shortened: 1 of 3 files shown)


lekt9:
yay thought it was a me problem where i vibed too hard

Issue #550, opened 2026-06-07
Enhancement: query-specific retrieval signals (call-graph distance to matched symbols; git co-change)


justrach:
## Context

codedb's `search` is already multi-signal: `searchContentRanked` (src/explore.zig:3135) ranks with **BM25/BM25+** (lines ~3230-3260) × **call-graph centrality** (`centralityBoost`, line 3286; built in `ensureCallGraph`, ~2957) × **path/test/doc penalties** (`pathRelevanceMultiplier`, ~2906-2944), and `context` blends BM25 × symbol-definition boost (src/mcp.zig). So this is *not* "add structure to ranking" — it's two signals that are genuinely absent and that target the hard case: **disambiguating the target file among many same-keyword hits.**

**Motivation:** in an ALMA-style retrieval experiment (engram re-ranking codedb features), the retrieval ceiling on a large repo (openclaw, 13.6k files) sat at **~0.30 MRR** — a *signal* limit: when a query identifier matches many files, the current features (lexical, **global** centrality, name-match, degree) can't tell which file a change actually targets. Both missing signals below are **query-specific**, unlike codedb's current global signals.

### 1. Query-specific call-graph distance
codedb builds the call graph and uses it for **global** centrality (per-file importance) — but ranking has no **query-specific** graph signal: *how close (in call/import hops) is this candidate to the symbols the query matched?* `findCallPath` (src/explore.zig:3087) exists but only for navigation, not ranking. Folding "graph distance to matched symbols" into the score would prefer files structurally near the query's definitions over distant same-keyword files.

### 2. Git co-change
codedb is git-history-blind: it reads only the HEAD SHA (src/git.zig, for snapshot invalidation) and uses file **mtime** as its only temporal signal — no commit-log / co-change. "Files historically changed together" is a strong signal for *which file a task touches*. Parsing `git log` into a co-change graph adds a high-value, query-relevant signal. (Needs real history; shallow clones won't have it.)

### (minor) Richer file-role
Role handling today is binary doc/code (`isDocLanguage`, src/explore.zig:191) + a heuristic test penalty (0.6x, ~2921). A richer multi-class role (config / generated / test / impl) used in ranking could route queries (a "config" query → `.toml`/`.json`; a "runtime" query → impl).

## Why now
All three extend machinery codedb already has (call graph, BM25, path heuristics) and attack the disambiguation ceiling that BM25 + *global* centrality can't break on its own.

_Surfaced via an ALMA-style retrieval experiment over codedb features._

[...]

PR #585, opened 2026-06-10
fix: memory/RSS + daemon-lifecycle bug hunt — 8 issues (#583 #584 #586 #587 #589 #592 #593 #594)

### LINKED ISSUE #583: index: disk-loaded word index never removes stale postings — re-index appends forever, deletes leave ghosts
## Problem

`WordIndex.readFromDisk` and `WordIndex.mmapFromDisk` both set `skip_file_words = true` (src/index.zig:790, :819), and `promoteIfBorrowed` keeps it that way. With `file_words` empty, `removeFile` hits `self.file_words.fetchRemove(path) orelse return` (src/index.zig:108) and becomes a **silent no-op** for every disk-loaded index — exactly the mode the warm daemon and CLI fast paths run in.

Consequences, all on the post-load write path (`indexFile` calls `removeFile` first, src/index.zig:186):

- **Re-indexing a file appends new postings while every stale one survives**: terms deleted from the file keep hitting it (at stale line numbers), and `(doc, line)` duplicates inflate BM25 term frequency.
- **Deleting a file leaves all of its postings live** — ghost hits with a valid-looking path.
- **Unbounded postings growth**: in a long-running watch/daemon session every file save grows the index; memory is never reclaimed (RSS). #400 fixed the `total_tokens`/`doc_lengths` counters in this mode but not the postings themselves.
- In pure zero-copy mmap mode `removeFile` is doubly a no-op: `path_to_id` is empty too, and unlike `indexFile` it never promotes the mmap to heap.

## Failing Test

`test_index.zig` — fails on current release tip:

```zig
test "issue-582: disk-loaded word index — re-index and removeFile must drop stale postings" {
    const alloc = testing.allocator;
    var wi = WordIndex.init(alloc);
    defer wi.deinit();
    try wi.indexFile("src/a.zig", "pub fn alphaToken() void {}\n");
    try wi.indexFile("src/b.zig", "pub fn betaToken() void {}\n");

    var tmp = testing.tmpDir(.{});
    defer tmp.cleanup();
    var path_buf: [std.fs.max_path_bytes]u8 = undefined;
    const dir_path_len = try tmp.dir.realPathFile(io, ".", &path_buf);
    const dir_path = path_buf[0..dir_path_len];
    try wi.writeToDisk(io, dir_path, null);

    // Heap fast-load: re-indexing a file must drop its old postings.
    var loaded = WordIndex.readFromDisk(io, dir_path, alloc).?;
    defer loaded.deinit();
    try loaded.indexFile("src/a.zig", "pub fn gammaToken() void {}\n");
    const stale = try loaded.searchDeduped("alphaToken", alloc);
    defer alloc.free(stale);
    try testing.expectEqual(@as(usize, 0), stale.len);
    const fresh = try loaded.searchDeduped("gammaToken", alloc);
    defer alloc.free(fresh);
    try testing.expectEqual(@as(usize, 1), fresh.len);

    // Deleting a file must drop its postings outright.
    loaded.removeFile("src/b.zig");
    const ghost = try loaded.searchDeduped("betaToken", alloc);
    defer alloc.free(ghost);
    try testing.expectEqual(@as(usize, 0), ghost.len);

    // Zero-copy mmap load: removeFile is a write — it must promote, not no-op.
    var mloaded = WordIndex.mmapFromDisk(io, dir_path, alloc).?;
    defer mloaded.deinit();
    mloaded.removeFile("src/a.zig");
    const mghost = try mloaded.searchDeduped("alphaToken", alloc);
    defer alloc.free(mghost);
    try testing.expectEqual(@as(usize, 0), mghost.len);
}
```

## Expected

After a disk fast-load, `indexFile` of changed content replaces a file's postings, and `removeFile` drops them — same observable behavior as a scratch-built index.

## Fix

Give `removeFile` a slow path for the no-`file_words` case: when `path_to_id` knows the path, sweep `index` for the doc_id (prune empty buckets), fix `doc_lengths`/`total_tokens`, blank + free the `id_to_path` slot (the skip-mode owner). In mmap mode, promote first (a remove is a write), but only if the path is actually tracked. The sweep is O(index) but runs at most once per file edit after a load; the fast path is untouched.

### LINKED ISSUE #584: hot_cache: ContentCache probe-window violations — overflow inserts land unreachable, holes break lookup and allow duplicate keys
## Problem

`ContentCache` is a bounded-probe (4-slot window) table, but three paths violate the window invariant (src/hot_cache.zig):

1. **Overflow eviction strands the new entry.** When all 4 window slots are occupied by other keys, `putImpl` evicts via `clockEvict()` — a *global* CLOCK hand that returns a slot anywhere in the table (L161). The new entry is written there, but `get` only probes the key's own 4-slot window: the entry can never be found. Its bytes (file contents, up to 64MB apiece) sit stranded in the cache while every lookup for that key misses and re-reads from disk. Under collision pressure the cache degrades to a byte-hoarding miss machine.
2. **Holes break lookup.** `get`/`remove` early-break on `!present and key_hash == 0` (L107, L195). A `remove` resets a slot to `empty_slot`, so an in-window entry stored *past* the removed slot becomes unreachable — stranded bytes + permanent misses until eviction happens to reap it.
3. **Duplicate keys.** `putImpl` takes the first empty slot in the window before checking the rest of it, so re-putting a key that lives past a hole inserts a second copy: double memory for the same path, inflated `count_`, and a stale shadow that serves old bytes if the front copy is ever removed.

All three are memory-hoarding/correctness bugs in the layer that fronts every content read.

## Failing Test

`test_core.zig` — fails on current release tip (scenario 1 trips first; 2 and 3 are also red individually):

```zig
test "issue-584: ContentCache probe-window — overflow inserts, holes, and duplicate keys" {
    const fnvBase = struct {
        fn base(key: []const u8, cap: u32) u32 {
            var h: u64 = 14695981039346656037;
            for (key) |b| {
                h ^= b;
                h *%= 1099511628211;
            }
            if (h == 0) h = 1;
            return @as(u32, @truncate(h)) % cap;
        }
    }.base;

    // Collect 5 keys sharing one probe-window base in [1, 60].
    var bufs: [5][16]u8 = undefined;
    var lens: [5]usize = undefined;
    var nkeys: usize = 0;
    var counts = [_]u8{0} ** 64;
    var pick: u32 = 0;
    var i: usize = 0;
    while (i < 4096) : (i += 1) {
        var tmp: [16]u8 = undefined;
        const k = std.fmt.bufPrint(&tmp, "k{d}", .{i}) catch unreachable;
        const b = fnvBase(k, 64);
        if (b < 1 or b > 60) continue;
        counts[b] += 1;
        if (counts[b] >= 5) {
            pick = b;
            break;
        }
    }
    try testing.expect(pick != 0);
    i = 0;
    while (i < 4096 and nkeys < 5) : (i += 1) {
        var tmp: [16]u8 = undefined;
        const k = std.fmt.bufPrint(&tmp, "k{d}", .{i}) catch unreachable;
        if (fnvBase(k, 64) == pick) {
            @memcpy(bufs[nkeys][0..k.len], k);
            lens[nkeys] = k.len;
            nkeys += 1;
        }
    }
    try testing.expectEqual(@as(usize, 5), nkeys);
    const k0 = bufs[0][0..lens[0]];
    const k1 = bufs[1][0..lens[1]];
    const k2 = bufs[2][0..lens[2]];
    const k3 = bufs[3][0..lens[3]];
    const k4 = bufs[4][0..lens[4]];

    // 1) Overflow insert must stay retrievable from its own probe window.
    {
        var cache = try ContentCache.initAlloc(testing.allocator, 64);
        defer cache.deinit();
        try cache.put(k0, "v0");
        try cache.put(k1, "v1");
        try cache.put(k2, "v2");
        try cache.put(k3, "v3");
        try cache.put(k4, "v4"); // window full -> eviction path
        try testing.expect(cache.get(k4) != null);
    }

    // 2) A hole left by remove() must not hide in-window entries behind it.
    // 3) Re-putting such an entry must not create a duplicate copy.
    {
        var cache = try ContentCache.initAlloc(testing.allocator, 64);
        defer cache.deinit();
        try cache.put(k0, "v0");
        try cache.put(k1, "v1");
        try cache.put(k2, "v2");
        cache.remove(k1); // hole between k0 and k2
        try testing.expect(cache.get(k2) != null);
        try cache.put(k2, "v2b");
        try testing.expectEqual(@as(u32, 2), cache.len());
    }
}
```

## Expected

Every entry the cache holds is retrievable via its probe window; a key occupies at most one slot; eviction frees the victim's bytes and the new entry replaces it *in-window*.

## Fix

Keep eviction inside the probe window: second-chance over the 4 window slots (first `!ref_bit` wins; if all are hot, clear them and take the base slot), free the victim, insert in place. Make `putImpl` scan the whole window for an existing key before taking the first empty slot, and drop the `key_hash == 0` early-breaks in `get`/`remove` (4 contiguous probes are cheap). The global `clockEvict`/`hand` machinery becomes dead and goes away.

### LINKED ISSUE #586: explore: symbol_index keys alias outline-owned strings — freed on re-index while shared-name entries survive (UAF)
## Problem

`rebuildSymbolIndexFor` keys the global `symbol_index` with `sym.name` slices owned by the file's outline (src/explore.zig:5081). When that file is re-indexed, `commitParsedFileOwnedOutline` deinits the prior outline (src/explore.zig:915), freeing the very bytes the map hashes and compares — but the map entry survives whenever **any other file shares the symbol name**, because `removeSymbolIndexFor` only drops entries whose location list becomes empty. In Zig code, `init`/`deinit`/`main` are shared by nearly every file, so the first edit of whichever file first inserted a shared name leaves a dangling key. The same happens on file deletion: `Explorer.removeFile` frees the outline (src/explore.zig:1513) after `removeSymbolIndexFor` kept shared-name entries alive.

Every subsequent lookup or iteration `eql()`s against freed memory — UB in release builds (with reused memory, potentially garbage names in `symbolSearch` results); under the DebugAllocator poison the entry becomes unreachable, silently degrading the O(1) index to the outline safety scans for the daemon's lifetime.

## Failing Test

`test_explore.zig` — fails on current release tip (`error.SymbolLostFromIndex`):

```zig
test "issue-586: symbol_index keys must survive re-index of the file that first inserted them" {
    var explorer = Explorer.init(testing.allocator, Explorer.DEFAULT_CONTENT_CACHE_CAPACITY);
    defer explorer.deinit();

    // a.zig inserts sharedFn first -> the map key aliases a.zig's outline string.
    try explorer.indexFile("src/a.zig", "pub fn sharedFn() void {}\n");
    try explorer.indexFile("src/b.zig", "pub fn sharedFn() void {}\n");

    // Re-index the key owner: its old outline is freed; the entry stays alive
    // because b.zig still holds a location.
    try explorer.indexFile("src/a.zig", "pub fn sharedFn() void {}\npub fn other() void {}\n");

    const locs = explorer.symbol_index.get("sharedFn") orelse return error.SymbolLostFromIndex;
    try testing.expect(locs.items.len >= 2);
}
```

## Expected

`symbol_index` entries stay valid and reachable regardless of which file's outline is replaced or removed.

## Fix

Make the map own its keys: dupe on first insert in `rebuildSymbolIndexFor`, free the key when `removeSymbolIndexFor` drops an emptied entry, and free keys in `Explorer.deinit`. One dupe per unique symbol name; uniform ownership also covers the snapshot fast-load case where names are borrowed from the mmap'd outline section.

### LINKED ISSUE #587: explore: removeFile leaves a dangling path key in skip_trigram_files
## Problem

`Explorer.removeFile` (src/explore.zig:1497) cleans `dep_graph`, `symbol_index`, `contents`, the word and trigram indexes, then frees the outlines key (src/explore.zig:1514) — but never removes the file's `skip_trigram_files` entry, whose key **aliases that freed outlines key**. Files enter `skip_trigram_files` via outline-only indexing (#507) and the `skip_trigram` path. After a delete, the tier-3 search scan iterates the dangling path and hands it to `readContentForSearch` — a freed-memory read (UB); with reused memory, an arbitrary string used as a file path.

## Failing Test

`test_explore.zig` — fails on current release tip (count stays 1):

```zig
test "issue-587: removeFile drops the path from skip_trigram_files" {
    var explorer = Explorer.init(testing.allocator, Explorer.DEFAULT_CONTENT_CACHE_CAPACITY);
    defer explorer.deinit();

    // Outline-only indexing registers the file in skip_trigram_files (#507).
    try explorer.indexFileOutlineOnly("src/gone.zig", "pub fn ghostFn() void {}\n");
    try testing.expectEqual(@as(usize, 1), explorer.skip_trigram_files.count());

    explorer.removeFile("src/gone.zig");
    try testing.expectEqual(@as(usize, 0), explorer.skip_trigram_files.count());
}
```

## Expected

After `removeFile`, the path is absent from every Explorer-side structure; no map retains a pointer into the freed outlines key.

## Fix

`_ = self.skip_trigram_files.remove(path);` in `Explorer.removeFile`, before the outlines entry is freed (keys alias the outlines key, so removal — not freeing — is the correct cleanup, matching the `commitParsedFileOwnedOutline` re-index path at src/explore.zig:892).

### LINKED ISSUE #589: watcher/snapshot: isSensitivePath misses OpenSSH default key names id_ecdsa, id_dsa, and *_sk FIDO variants
## Problem

The `sensitive_names` exact list in both `isSensitivePath` copies (src/watcher.zig:1200, src/snapshot.zig:1328) covers `id_rsa` and `id_ed25519` but misses the other ssh-keygen default basenames: `id_ecdsa`, `id_dsa`, and the FIDO variants `id_ecdsa_sk` / `id_ed25519_sk`. Inside `~/.ssh/` the directory rule catches them, but a key copied into a repo — `deploy/id_ecdsa` — is indexed and readable while `deploy/id_rsa` is blocked. Same class as the #585-adjacent secrets pass (`*.env` suffix, `.git-credentials`).

## Failing Test

`test_mcp.zig` — fails on current release tip:

```zig
test "issue-588: isSensitivePath blocks all OpenSSH default private key names" {
    const keys = [_][]const u8{ "id_ecdsa", "id_dsa", "id_ecdsa_sk", "id_ed25519_sk" };
    for (keys) |k| {
        try testing.expect(watcher.isSensitivePath(k));
        try testing.expect(snapshot_mod.isSensitivePath(k));
    }
    try testing.expect(watcher.isSensitivePath("deploy/id_ecdsa"));
    // negatives — names that merely start like a key must stay indexable
    try testing.expect(!watcher.isSensitivePath("id_map.zig"));
    try testing.expect(!watcher.isSensitivePath("identity.ts"));
}
```

## Expected

All six ssh-keygen default private-key basenames are blocked everywhere, matching the existing `id_rsa`/`id_ed25519` handling.

## Fix

Add the four names to the exact list in both copies (they all start with `i`, so the fast-path branch already routes them to the exact-match check). The #528 parity test keeps the copies honest.

### LINKED ISSUE #592: cli-daemon: spawn race under concurrent CLI calls — calls fork duplicate daemons, each rescans the index, orphans churn CPU afterward
**Version:** codedb 0.2.5824, macOS (Darwin 25.5.0), repo: openclaw (~19.7k files)

## Summary

When many `codedb <root> <cmd>` CLI processes start concurrently and the daemon is not fully settled, several of them fork their **own** `cli-daemon` instead of connecting to the existing one. Each duplicate daemon re-scans/rebuilds the index, the concurrent calls block behind those rebuilds, and the duplicates **stay alive afterward** (PPID 1), churning CPU and destabilizing later calls — which retriggers the race on the next burst.

## Measured

engram runs batches of independent `deps`/`search`/`glob` calls through one `/bin/sh` with N background lanes against a warm daemon:

- N = 16 concurrent: stable — a 15-query benchmark runs in ~0.45s, daemon count stays at 1.
- N = 32 concurrent: stampede — same benchmark takes **22s** (reproducible: 22.1/22.7/23.4s), and `ps` during the run shows the daemon count climbing (observed 3 → 7 within one run; at one point 9 orphaned `cli-daemon` processes for the same root, ages 1–3 min, each at 10–23% CPU long after their spawning calls exited).

Snapshot mid-stampede (all same root):

```
86330   11.7%  codedb /Users/.../openclaw cli-daemon
86519    1.5%  codedb /Users/.../openclaw cli-daemon
86622    0.0%  codedb /Users/.../openclaw cli-daemon
86957   15.7%  codedb /Users/.../openclaw cli-daemon
86963   10.8%  codedb /Users/.../openclaw cli-daemon
...        (PPID 1 — orphaned; spawning CLI calls long gone)
```

Once poisoned, even single serial calls intermittently pay a full ~15s rebuild until the orphans are killed (`pkill -f cli-daemon`) and one fresh daemon settles.

## Expected

Daemon spawn should be mutually exclusive per project (lockfile / atomic socket bind): exactly one daemon ever spawns, racing callers wait on the winner, and a daemon that loses the race exits instead of lingering.

## Workaround

We cap engram's batch concurrency at 16 and treat it as a hard ceiling, but the safe number is presumably machine/repo dependent — a proper spawn lock would remove the cliff entirely.

### LINKED ISSUE #593: index: mmap trigram index — removeFile silently no-ops; overlay promotion never masks stale base entries
## Problem

`AnyTrigramIndex.removeFile` is `{}` in pure-mmap mode (src/index.zig:2518), and the `mmap_overlay` promotion keeps querying the base for paths the overlay has since superseded:

- a file deleted while the index is zero-copy stays `containsFile == true` forever, its trigrams keep producing candidates, and `fileCount` stays inflated
- a file edited after an mmap load feeds candidates from **both** its old content (mmap base) and its new content (overlay) — `containsFile` ORs the two (src/index.zig:2494) and `candidates`/`candidatesRegex` merge both sides with no masking

Ghost candidates are re-verified against real content downstream, so the damage is wasted I/O per search plus wrong containment/count answers — but `containsFile` gates the tier-3 supplemental scan (src/explore.zig:3531, :5294), so a stale `true` can silently drop a re-indexed file out of the scan tiers that were supposed to cover it.

## Failing Test

`test_index.zig` — fails on current release tip (first assert: containsFile stays true after removeFile):

```zig
test "issue-590: mmap trigram index — removeFile takes effect and re-index masks stale base entries" {
    defer arena.deinit();
    const allocator = arena.allocator();

    var explorer = Explorer.init(testing.allocator, Explorer.DEFAULT_CONTENT_CACHE_CAPACITY);
    defer explorer.deinit();
    try explorer.indexFile("src/auth.zig", "pub fn handleAuth(req: *Request) !void { validate(req); }");
    try explorer.indexFile("src/gate.zig", "pub fn checkGate(ctx: *Context) !bool { return ctx.authenticated; }");
    try explorer.indexFile("src/util.zig", "pub fn formatStr(buf: []u8, args: anytype) !void {}");

    var tmp_dir = testing.tmpDir(.{});
    defer tmp_dir.cleanup();
    var path_buf: [std.fs.max_path_bytes]u8 = undefined;
    const tmp_path_len = try tmp_dir.dir.realPathFile(io, ".", &path_buf);
    const tmp_path = path_buf[0..tmp_path_len];
    try explorer.trigram_index.writeToDisk(io, tmp_path, null);

    const mmap_idx = MmapTrigramIndex.initFromDisk(io, tmp_path, testing.allocator) orelse
        return error.MmapInitFailed;
    var any_idx = AnyTrigramIndex{ .mmap = mmap_idx };
    defer any_idx.deinit();

    // A delete while zero-copy must take effect, not silently no-op.
    any_idx.removeFile("src/gate.zig");
    try testing.expect(!any_idx.containsFile("src/gate.zig"));
    if (any_idx.candidates("checkGate", allocator)) |cands| {
        for (cands) |p| try testing.expect(!std.mem.eql(u8, p, "src/gate.zig"));
    }

    // Re-indexing must mask the base's stale trigrams for that path.
    try any_idx.indexFile("src/auth.zig", "pub fn renamedAuth() void {}");
    if (any_idx.candidates("handleAuth", allocator)) |cands| {
        for (cands) |p| try testing.expect(!std.mem.eql(u8, p, "src/auth.zig"));
    }
    const fresh = any_idx.candidates("renamedAuth", allocator) orelse return error.NoCandidates;
    var found = false;
    for (fresh) |p| {
        if (std.mem.eql(u8, p, "src/auth.zig")) found = true;
    }
    try testing.expect(found);

    // File accounting follows: 3 on disk, one removed.
    try testing.expectEqual(@as(u32, 2), any_idx.fileCount());
}
```

## Expected

Removal and re-index behave identically across heap, mmap, and overlay modes: base entries for superseded/removed paths stop answering.

## Fix

Add a `masked` path set (owned keys) to `MmapOverlay`: `indexFile`/`removeFile` mask the path (removeFile on `.mmap` promotes to an overlay first — a remove is a write), `containsFile` and the `candidates`/`candidatesRegex` merges filter base hits through it, and `fileCount` subtracts a maintained masked-in-base counter.

### LINKED ISSUE #594: explore: indexFile OOM paths — per-parse temp Explorer panics the process; failure-restore reads freed prior_content
## Problem

Two defects on `indexFile`'s failure path, found by sweeping a fail-once allocator across a re-index:

1. **Every `indexFile` builds a throwaway full `Explorer`** (`parseContentForIndexing`, src/explore.zig:1307) just to carry an allocator into the line parsers — allocating and memsetting a 4096-slot ContentCache (~330KB) per parsed file. Worse, `ContentCache.init` **panics** on allocation failure, so an OOM during routine indexing aborts the daemon instead of returning `error.OutOfMemory`. The failing test below dies with `panic: ContentCache.init: OOM allocating 4096 slots` before it can even exercise defect 2.

2. **`commitParsedFileOwnedOutline` reads `prior_content` out of the content cache and then calls `contents.put`** (src/explore.zig:862-863), which frees that value in place. The trigram-failure `errdefer` (src/explore.zig:876) later "restores" the word index by re-indexing `prior_content` — freed memory. Under the DebugAllocator poison the tokenizer finds no words, so a failed re-index silently wipes the file's postings instead of restoring them; release builds tokenize whatever the allocator reused the bytes for.

## Failing Test

`test_explore.zig` — on current release tip the suite binary **aborts** (defect 1); with the panic fixed it fails `error.PriorContentLostOnFailedReindex` (defect 2):

```zig
test "issue-591: failed re-index restores the word index from valid prior content" {
    const FailOnce = struct {
        backing: std.mem.Allocator,
        next_index: usize = 0,
        fail_index: usize,
        // fail exactly the Nth allocation, then recover — std.testing's
        // FailingAllocator fails forever from fail_index on, which would also
        // starve the errdefer restore we are trying to verify
        ... (vtable forwarding to backing, alloc fails once at fail_index) ...
    };

    // Pass 1: count the allocation window of the re-index on a successful run.
    // Pass 2: fail each allocation in that window exactly once; whenever the
    // re-index errors, the OLD content must still be word-searchable:
    //     if (explorer.indexFile("src/a.zig", "pub fn betaTok() void {}\n")) |_| {
    //         try testing.expect(explorer.word_index.search("betatok").len > 0);
    //     } else |_| {
    //         if (explorer.word_index.search("alphatok").len == 0)
    //             return error.PriorContentLostOnFailedReindex;
    //     }
}
```

## Expected

OOM during indexing is an error, not a process abort; a failed re-index leaves the word index reflecting the old content (the restore reads live bytes).

## Fix

Split `Explorer.init` into a fallible `initFallible` (used by `init`, which keeps its panic for real explorers) and give `parseContentForIndexing` a 1-slot parser shell via `try initFallible(allocator, 1)` — the parsers only use `self.allocator`. Move `contents.put` after the word/trigram updates in `commitParsedFileOwnedOutline` so `prior_content` stays valid through the entire fallible region the errdefer protects.

justrach:
Memory/RSS-focused bug hunt grown across three rounds, all red→green where a red state exists. Suite **761/761**.

## Round 1 — index growth & cache integrity
- **#583** — disk-loaded word index never removes stale postings: bulk paths (`readFromDisk`/`mmapFromDisk`/`mergeShard` — the warm-daemon config) left `removeFile` a no-op → ghost/stale postings, doubled BM25 tf, unbounded growth per save. Fix: doc_id sweep slow path, hybrid per-file word lists, `writeToDisk` file table from `id_to_path`, mmap promote-on-remove.
- **#584** — ContentCache probe-window violations: global-CLOCK overflow eviction stranded entries outside the lookup window (bytes hoarded, permanent misses, disk re-reads); holes broke lookups; duplicate keys possible. Fix: full-window scan, in-window second-chance eviction, PROBE_LIMIT 4→8, dead CLOCK machinery removed. Also explains + kills the longstanding ~1-in-6 snapshot-freshness flake (see PR comment).

## Round 2 — dangling references & secrets
- **#586** — `symbol_index` keyed by outline-owned `sym.name`: re-index/delete of the first inserter freed the key while shared-name entries survived (`init`/`deinit`/`main` are shared by nearly every Zig file) → UAF on every probe, names silently dropping to slow scans. Fix: map owns its keys.
- **#587** — `Explorer.removeFile` left a dangling path in `skip_trigram_files` (key aliases the freed outlines key) → tier-3 scan iterated freed memory. Fix: remove alongside the other structures.
- **#589** — `isSensitivePath` missed `id_ecdsa`/`id_dsa`/`*_sk` SSH keys: `deploy/id_ecdsa` indexed while `deploy/id_rsa` blocked. Both copies fixed.

## Round 3 — mmap trigram ghosts + the indexFile OOM family + daemon stampede
- **#593** — `AnyTrigramIndex.removeFile` no-op'd in pure-mmap mode and the overlay never masked base entries: deleted files stayed "contained" with live candidates; edited files answered from both old and new content. Fix: `MmapOverlay.masked` path set (owned keys), filtered through `containsFile`/`candidates`/`candidatesRegex` (shared merge helper), `fileCount` correction, promote-on-remove.
- **#594** — a fail-once-allocator sweep over one re-index surfaced **six** failure-path defects, all fixed: per-parse throwaway full Explorer whose `ContentCache.init` *panics* on OOM (now a fallible 1-slot parser shell); `errdefer`+`defer` double-deinit of the parsed outline; the `prior_content` UAF (cache `put` now last fallible step + one function-scoped word-index restore); restore gap for failures inside `word_index.indexFile`; orphaned `getOrPut` entries on key-dupe failure in `indexOneToken`/`mergeShard` (stack-buffer keys, undefined values → segfault); `removeSymbolIndexFor` poisoning the in-map value before a fallible append.
- **#592** — cli-daemon spawn stampede (filed from the engram side with measurements): concurrent cold calls forked a daemon each, duplicates paid full rescans, and the stale-socket unlink let late arrivals steal the winner's live socket → orphans churning CPU. Fix: per-project exclusive flock taken before the load (kernel-released on exit, crash-safe) + CLI-side spawn probe. Verified live: 8 concurrent cold calls → exactly 1 daemon. The related #591 (cold path answers from the wrong repo, 0.2.5824) no longer reproduces on this branch — closed with repro evidence.

## Also filed (enhancements, per #550/#564 precedent)
- #596 — byte-budgeted ContentCache eviction (entry-bounded today, byte-blind)
- #597 — Store hardening: advisory-lock failure proceeds unlocked; data_log never compacts

## Notes
- Found by dogfooding `codedb` (outline/deps/context/status) against its own tree plus targeted reads and a fail-once allocator sweep.
- Known + parked: `commitParsedFileOwnedOutline` metadata flags drift on failure paths (cosmetic); serve-vs-cli-daemon socket handover quirk (pre-existing, proxy-disabled fallback).
- The `test_mcp` issue-148 EOF test and the three `test_index` perf-threshold tests flake when their binaries re-run under concurrent load; all pass consistently on green suites.

Closes #583. Closes #584. Closes #586. Closes #587. Closes #589. Closes #592. Closes #593. Closes #594.

### DIFF
--- a/src/explore.zig
+++ b/src/explore.zig
@@ -749,6 +749,7 @@ pub const Explorer = struct {
 
         var sym_iter = self.symbol_index.iterator();
         while (sym_iter.next()) |entry| {
+            self.allocator.free(entry.key_ptr.*);
             entry.value_ptr.deinit(self.allocator);
         }
         self.symbol_index.deinit();
@@ -1504,6 +1505,7 @@ pub const Explorer = struct {
         }
         self.dep_graph.remove(path);
         self.removeSymbolIndexFor(path);
+        _ = self.skip_trigram_files.remove(path);
         self.contents.remove(path);
         self.word_index.removeFile(path);
         self.trigram_index.removeFile(path);
@@ -2252,16 +2254,16 @@ pub const Explorer = struct {
     /// name-only resolution. Conservative: only the highest-collision names.
     fn isUbiquitousName(name: []const u8) bool {
         const common = [_][]const u8{
-            "init",        "deinit",   "append",   "get",       "set",
-            "put",         "getOrPut", "remove",   "contains",  "has",
-            "count",       "len",      "items",    "alloc",     "free",
-            "create",      "destroy",  "lock",     "unlock",    "tryLock",
-            "next",        "iterator", "clone",    "dupe",      "slice",
-            "reset",       "clear",    "format",   "hash",      "eql",
-            "write",       "writeAll", "print",    "read",      "close",
-            "open",        "value",    "key",      "toOwnedSlice", "pop",
-            "push",        "find",     "add",      "new",       "build",
-            "run",         "deinit",   "toString", "valueOf",   "of",
+            "init",   "deinit",   "append",   "get",          "set",
+            "put",    "getOrPut", "remove",   "contains",     "has",
+            "count",  "len",      "items",    "alloc",        "free",
+            "create", "destroy",  "lock",     "unlock",       "tryLock",
+            "next",   "iterator", "clone",    "dupe",         "slice",
+            "reset",  "clear",    "format",   "hash",         "eql",
+            "write",  "writeAll", "print",    "read",         "close",
+            "open",   "value",    "key",      "toOwnedSlice", "pop",
+            "push",   "find",     "add",      "new",          "build",
+            "run",    "deinit",   "toString", "valueOf",      "of",
         };
         for (common) |c| if (std.mem.eql(u8, name, c)) return true;
         return false;
@@ -4132,7 +4134,7 @@ pub const Explorer = struct {
                 try appendOutlineSymbol(a, outline, name, kind, line_num, line);
             }
         }
-        if (startsWith(line, "import ") or containsAny(line, &.{ "require(" })) {
+        if (startsWith(line, "import ") or containsAny(line, &.{"require("})) {
             try appendOutlineSymbol(a, outline, line, .import, line_num, null);
             if (extractStringLiteral(line)) |path| {
                 try appendImportPath(a, outline, path);
@@ -5080,6 +5082,13 @@ pub const Explorer = struct {
         for (outline.symbols.items) |sym| {
             const gop = self.symbol_index.getOrPut(sym.name) catch continue;
             if (!gop.found_existing) {
+                // The map owns its keys: sym.name belongs to the outline and
+                // dies with it on re-index, while shared-name entries survive.
+                const owned = self.allocator.dupe(u8, sym.name) catch {
+                    _ = self.symbol_index.remove(sym.name);
+                    continue;
+                };
+                gop.key_ptr.* = owned;
                 gop.value_ptr.* = std.ArrayList(SymbolLocation).empty;
             }
             gop.value_ptr.append(self.allocator, .{
@@ -5112,7 +5121,9 @@ pub const Explorer = struct {
             }
         }
         for (to_remove.items) |key| {
-            _ = self.symbol_index.remove(key);
+            if (self.symbol_index.fetchRemove(key)) |kv| {
+                self.allocator.free(kv.key);
+            }
         }
     }
 
--- a/src/explore.zig
+++ b/src/explore.zig
@@ -724,10 +724,15 @@ pub const Explorer = struct {
     }
 
     pub fn init(allocator: std.mem.Allocator, content_cache_capacity: u32) Explorer {
+        return initFallible(allocator, content_cache_capacity) catch
+            std.debug.panic("Explorer.init: OOM allocating {d} content cache slots", .{content_cache_capacity});
+    }
+
+    fn initFallible(allocator: std.mem.Allocator, content_cache_capacity: u32) !Explorer {
         return .{
             .outlines = std.StringHashMap(FileOutline).init(allocator),
             .dep_graph = DependencyGraph.init(allocator),
-            .contents = ContentCache.init(allocator, content_cache_capacity),
+            .contents = try ContentCache.initAlloc(allocator, content_cache_capacity),
             .symbol_index = std.StringHashMap(std.ArrayList(SymbolLocation)).init(allocator),
             .symbol_index_complete = true,
             .word_index = WordIndex.init(allocator),
@@ -828,8 +833,12 @@ pub const Explorer = struct {
 
     pub fn commitParsedFileOwnedOutline(self: *Explorer, path: []const u8, content: []const u8, outline: FileOutline, full_index: bool, skip_trigram: bool) !void {
         var owned_outline = outline;
-        errdefer owned_outline.deinit();
-        var persistent_outline = try cloneOutline(&owned_outline, self.allocator);
+        // One deinit only: an errdefer here would stack with the defer below
+        // and double-free the parsed outline on any post-clone error.
+        var persistent_outline = cloneOutline(&owned_outline, self.allocator) catch |err| {
+            owned_outline.deinit();
+            return err;
+        };
         defer owned_outline.deinit();
         errdefer persistent_outline.deinit();
         if (persistent_outline.owns_path) {
@@ -861,7 +870,17 @@ pub const Explorer = struct {
         persistent_outline.path = stable_path;
 
         const prior_content = self.contents.get(stable_path);
-        try self.contents.put(stable_path, content);
+        // Any failure below must put the word index back the way it was —
+        // indexFile wipes the old postings before adding new ones, and search
+        // serves whatever the postings say. prior_content stays valid through
+        // every fallible step because contents.put runs last.
+        errdefer if (full_index and !self.defer_word_index) {
+            if (prior_content) |old| {
+                self.word_index.indexFile(stable_path, old) catch {};
+            } else {
+                self.word_index.removeFile(stable_path);
+            }
+        };
 
         if (full_index) {
             // A disabled word index (cold CLI scan keeps it off to save memory)
@@ -872,13 +891,6 @@ pub const Explorer = struct {
                 self.word_index_can_load_from_disk = false;
             }
             if (!self.defer_word_index) try self.word_index.indexFile(stable_path, content);
-            // If trigram indexing fails below, restore word_index to its previous state
-            // to prevent word_index and trigram_index from diverging.
-            errdefer if (prior_content) |old| {
-                self.word_index.indexFile(stable_path, old) catch {};
-            } else {
-                self.word_index.removeFile(stable_path);
-            };
             if (self.word_index_complete) {
                 self.word_index_generation +%= 1;
             }
@@ -911,6 +923,10 @@ pub const Explorer = struct {
 
         try self.rebuildDepsFor(stable_path, &persistent_outline);
         self.rebuildSymbolIndexFor(stable_path, &persistent_outline, !is_new);
+
+        // Last fallible step: put frees the prior cache value in place, so it
+        // must run only once nothing after it can still need prior_content.
+        try self.contents.put(stable_path, content);
 
         outline_gop.value_ptr.* = persistent_outline;
         if (prior_outline) |*old_outline| old_outline.deinit();
@@ -1304,7 +1320,10 @@ pub const Explorer = struct {
     }
 
     pub fn parseContentForIndexing(allocator: std.mem.Allocator, path: []const u8, content: []const u8) !ParsedFile {
-        var parser = Explorer.init(allocator, DEFAULT_CONTENT_CACHE_CAPACITY);
+        // A parser shell only carries the allocator into the parse methods: a
+        // 1-slot cache keeps the per-file cost flat, and OOM here must surface
+        // as an error, not ContentCache.init's daemon-killing panic.
+        var parser = try Explorer.initFallible(allocator, 1);
         defer parser.deinit();
         var parsed_outline = try parseOutlineWithParser(&parser, path, content);
         defer parsed_outline.deinit();
@@ -5116,12 +5135,16 @@ pub const Explorer = struct {
                 }
             }
             if (list.items.len == 0) {
-                list.deinit(self.allocator);
+                // Deinit only at actual removal below: ArrayList.deinit poisons
+                // the in-map value, so a failed append here must leave a VALID
+                // (empty) list behind, not a landmine for the next iteration.
                 to_remove.append(self.allocator, entry.key_ptr.*) catch {};
             }
         }
         for (to_remove.items) |key| {
             if (self.symbol_index.fetchRemove(key)) |kv| {
+                var hits = kv.value;
+                hits.deinit(self.allocator);
                 self.allocator.free(kv.key);
             }
         }
--- a/src/hot_cache.zig
+++ b/src/hot_cache.zig
@@ -11,14 +11,15 @@ const std = @import("std");
 pub const ContentCache = struct {
     slots: []Slot,
     capacity: u32,
-    hand: u32,
     count_: u32,
     allocator: std.mem.Allocator,
     hits_: std.atomic.Value(u64),
     misses_: std.atomic.Value(u64),
     evictions_: std.atomic.Value(u64),
 
-    const PROBE_LIMIT: u32 = 4;
+    // 8-way set-associative: at typical occupancy a full window (the only
+    // eviction trigger) is vanishingly rare; probes stay short and contiguous.
+    const PROBE_LIMIT: u32 = 8;
 
     pub const Slot = struct {
         key_hash: u64,
@@ -57,7 +58,6 @@ pub const ContentCache = struct {
         return .{
             .slots = slots,
             .capacity = capacity,
-            .hand = 0,
             .count_ = 0,
             .allocator = allocator,
             .hits_ = std.atomic.Value(u64).init(0),
@@ -73,7 +73,6 @@ pub const ContentCache = struct {
         return .{
             .slots = slots,
             .capacity = capacity,
-            .hand = 0,
             .count_ = 0,
             .allocator = allocator,
             .hits_ = std.atomic.Value(u64).init(0),
@@ -104,7 +103,6 @@ pub const ContentCache = struct {
                 _ = self.hits_.fetchAdd(1, .monotonic);
                 return slot.value;
             }
-            if (!slot.present and slot.key_hash == 0) break;
         }
         _ = self.misses_.fetchAdd(1, .monotonic);
         return null;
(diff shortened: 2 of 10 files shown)


justrach:
One more datapoint: the longstanding flake in `test_snapshot "parallel freshness load re-indexes changed files, restores the rest"` (fails ~1 in 6 runs with `error.MissingContent`) is a live manifestation of #584. The freshness loader keys the ContentCache by **absolute** paths, so the tmpdir prefix re-rolls the FNV hashes every run; with 1002 keys in 16384 slots and a 4-slot probe window, the expected number of full windows per run is ~0.2 — and on the runs where one filled, the old overflow path stranded the entry where `get()` could never find it → `contents.get(abs_path)` → `MissingContent`. With in-window eviction + the 8-way window, a full window at that occupancy is ~10⁻⁶ per run. The flake should be gone with this PR.

justrach (commented):
Automated review (Sonnet pass + verification), defects-only scope:

**Verdict: safe to merge.** All 8 filed bugs verified genuinely closed; no new memory-safety defects, double-frees, or UAFs found in the bundle.

Two low-severity, non-blocking observations:

1. **#592 spawn-race fix is a window-shrink, not full TOCTOU elimination** — `daemonLockAvailable` releases the probe lock before the fork, so under a true stampede N callers can still fork N daemons that race `daemonLockTryAcquire`; losers exit before the expensive load. Outcome (no duplicate *running* daemons, no rescan churn) matches the issue, just with N−1 wasted forks in the worst case. The comment's 'losers simply cold-serve' slightly undersells that they spawn-and-exit.

2. **`MmapOverlay.mask()` swallows OOM silently** — if the `dupe`/`put` fails, the base entry stays unmasked and `containsFile` keeps answering true for stale content for the duration of memory pressure (same degraded behavior as pre-fix, so not a regression). Worth a follow-up: log the swallow or propagate so `indexFile` can retry.

Cleared after explicit verification (so nobody re-checks): #594 prior_content UAF ordering and owned_outline error-path deinit; #587 skip_trigram_files key lifetime (remove-before-free, value-equality lookup); #586 symbol_index key dup/free symmetry; #583 skip_file_words separate allocations (no double-free); #584 ContentCache sentinel removal correctness for the 8-probe window; mergeOverlayCandidates frees on all paths; isSensitivePath updated in both copies (snapshot+watcher; id_rsa_sk omission is correct — no such OpenSSH name); parseContentForIndexing capacity=1 harmless; word_index_generation bump on failed trigram = extra-but-valid disk write.

PR #622, opened 2026-06-16
feat(parser): add OCaml (.ml / .mli) support

Macho0x:
Adds OCaml language support to the line-based parser.

- Detect `.ml` / `.mli` files
- Parse `open` / `include` / `module` / `type` / `external` / `exception` / `val` / `let` / `and`
- Handle nested `(* ... *)` comments
- Classify `let` / `and` as function when `->` or `fun` / `function` is present
- Add parser test

`zig build test-parser`: 70/70 passed.


justrach:
Merged manually into release/0.2.5833-mcp2026 as 207da6c. One trivial append-append conflict in src/test_parser.zig resolved (kept both the rescript and ocaml tests); full `zig build test` including `test-parser` green on the merged tree.

Issue #623, opened 2026-06-17
tools: agents default to search+read+edit and skip the structural tools (symbol/callers/deps/outline)


justrach:
## Summary
Agent trajectories using codedb tools show **high length variance**, and on a minority of (large-repo) tasks the agent **fails to converge and burns 3-5× the tokens** of the native-tools baseline. The behavior is **intermittent** - the same task can run clean once and run away the next time.

## Evidence (SWE-bench Lite, Sonnet 4.6, codedb vs native, total tokens / turns)
| instance | native | codedb | codedb turns |
|---|--:|--:|--:|
| sympy__sympy-12419 | 978k | **3,362k** (-244%) | 50 (vs 28) |
| django__django-11910 | 242k | **1,328k** (-448%) | 30 (vs 9) |
| django__django-11133 | 270k | 572k (run A) / **clean ~9 calls (run B)** | 19 / 9 |

On 44/51 tasks codedb used fewer tokens (median −34%), but these tail runaways drag the aggregate down and flip the cost delta to roughly break-even. Re-running the same instance shows the runaway is non-deterministic.

## Impact
A handful of non-convergent runs erase the otherwise-solid efficiency win, and make worst-case cost/latency unpredictable.

## Suggested direction
- Loop/redundancy detection: same `search` query repeated, same file `read` repeatedly, edit→re-read→edit churn.
- A soft convergence signal/guardrail the host agent can act on.

_Found via an independent SWE-bench Lite token-efficiency benchmark: identical agent (\`claude -p\`, Sonnet 4.6) and tasks, only the tool surface differs - native Read/Grep/Edit vs codedb MCP tools. N=51 paired instances; full harness + data available._

justrach:
Duplicate of #626 (identical title, same root cause: agents default to `search → read → edit` and skip the structural tools).

The structural-steering fix tracked under #626 has landed: prescriptive tool descriptions + strengthened server instructions, plus two in-context runtime nudges — a bare-identifier `search` now points at `codedb_symbol`/`codedb_callers`, and a whole-file `read` (≥400 lines, no range) points at `codedb_outline`. Verified live: the nudge fires through the MCP server, and a Sonnet 4.6 agent reads the new server instruction and cites it when choosing structural tools first.

[...]

PR #627, opened 2026-06-18
feat(mcp): steer agents to structural tools (#626)

### LINKED ISSUE #626: tools: agents default to search+read+edit and skip the structural tools (symbol/callers/deps/outline)
## Summary
When the codedb MCP tools are available, agents almost exclusively use **`search` → `read` → `edit`** (a grep+cat+patch flow) and effectively **never invoke the structural-intelligence tools** - `symbol`, `callers`, `callpath`, `deps`, `outline`, `context`. codedb's actual differentiator (the code graph) goes unexercised; in practice it's used like ripgrep with smaller output.

## Evidence
Tool-call traces of agent runs over SWE-bench Lite (django/sympy/requests/etc.). Representative full trace for one task:
```
ToolSearch -> search("def content") -> read(response.py) -> search("def make_bytes")
 -> read(response.py) -> ToolSearch -> edit -> edit -> edit
```
Across audited runs the pattern is the same: `search`/`read`/`edit` only. No `symbol`/`callers`/`deps`/`outline`/`context` calls were observed.

## Impact
- The measured token savings vs native tools (~34% median) come from **compact tool *output***, not from smarter graph-based navigation - i.e. codedb is winning as "a leaner grep/read", leaving its headline value on the table.
- Grep-style exploration on large repos correlates with the runaway trajectories tracked separately.

## Suggested direction
- Strengthen tool **descriptions / server instructions** to steer agents to structural tools first (e.g. "to locate a definition use `symbol`, not `search`; to find usages use `callers`").
- Or consolidate the generic `search`/`read` so the structural path is the path of least resistance.

_Found via an independent SWE-bench Lite token-efficiency benchmark: identical agent (\`claude -p\`, Sonnet 4.6) and tasks, only the tool surface differs - native Read/Grep/Edit vs codedb MCP tools. N=51 paired instances; full harness + data available._

justrach:
## Problem

Agents on the codedb MCP surface default to `search → read → edit` and skip the structural tools (`symbol`/`callers`/`deps`/`outline`), so the code graph goes unexercised — codedb ends up used like a leaner grep. (See #626.)

## Change

Make the structural path the path of least resistance, two ways:

- **Descriptions / server instructions** — reframed to prescribe structural tools first and cast `codedb_search` as a substring/phrase *fallback*, not the default.
- **Runtime nudges** (the part that fights the prior in-context):
  - `codedb_search` on a bare identifier that resolves to an indexed symbol prepends a one-line pointer to `codedb_symbol`/`codedb_callers` (text output only; skipped for `format=json`).
  - `codedb_read` of a whole file (≥400 lines, no range) prepends a pointer to `codedb_outline`; wired into **both** the cached (`renderReadBytes`) and uncached read paths.

## Tests

`src/test_mcp.zig` — `issue-626` cases cover the gating logic: `isBareIdentifier` (fires on `make_bytes`, not on `def content` / `make_bytes(` / `2fast`) and `fullFileReadHint` (null under 400 lines, fires above). Full suite green: `zig build test` → exit 0.

## Verification

Driven live through a fresh `codedb mcp` server: the search nudge fires for a bare indexed identifier and is correctly **absent** from the CLI path and from `format=json`. A Sonnet 4.6 agent reads the new server instruction and cites it when choosing structural tools first.

Caveat worth stating: Sonnet 4.6 in Claude Code already navigates structurally (outline-before-read, ranged reads, `symbol`/`callers` by reflex), so on these tasks neither runtime nudge actually *triggered* — the steering is confirmation/insurance for this model. The payoff is on grep-happier agent loops; a SWE-bench-Lite A/B is the test that can show the nudges deflecting a bad call, and is tracked separately.

## Issues

Closes #626. #623 was closed as a duplicate; its distinct concern (tail-runaway / loop-redundancy convergence guardrail) is **not** addressed here and was preserved in that issue's close comment.


justrach:
Already on release/0.2.5826: both #626 commits (structural-tool steering + codedb_deps surfacing) are patch-equivalent on the release tip. Discarding in favor of the tip.

PR #630, opened 2026-06-19
fix: absolute in-project paths (#629) + regex alternation false negatives (#628)

### LINKED ISSUE #628: regex search: `|` alternation silently returns 0 matches (false negatives)
## Summary

In `search` / `faster_search` with `mode=regex` (flavor `zigrep_basic`), the `|` alternation operator does **not** work and the search returns **0 matches silently** — no error, no warning. This reads as a confident "pattern not found" when the terms are actually present, which is actively misleading and caused me to draw wrong conclusions while searching a codebase.

## Reproduction

File `index.ts` line 1:
```ts
import { generateText, streamText, createGateway } from "ai";
```

| pattern | mode | result | expected |
|---|---|---|---|
| `createGateway` | regex | 2 matches ✓ | 2 |
| `generateText\|streamText` | regex | **0 matches** ✗ | matches line 1 |
| `nonexistentword\|createGateway` | regex | **0 matches** ✗ | matches `createGateway` |

The single-term regex finds it; the moment an alternation `|` is added, the whole query returns 0 — even though one of the alternatives clearly matches.

## Impact

- Silent false negatives. A multi-term `a|b|c` "does any of these exist" sweep — a very common first-pass search — returns 0 and looks authoritative.
- No signal that alternation is unsupported. If `zigrep_basic` is intentionally POSIX BRE (where `|` is a literal char, alternation requires `\|`), that's a reasonable engine choice, but it's a footgun without feedback.

## Suggested fix (any one)

1. Support ERE-style `|` alternation in `mode=regex`, **or**
2. If `|` is literal in BRE, detect a bare `|` in a regex pattern and surface a hint (e.g. `"| is literal in basic regex; use \\| for alternation or split into multiple queries"`), **or**
3. At minimum, document the flavor's alternation semantics in the tool description.

## Environment

- codedbpro v0.2.9
- Both `search` and `faster_search` regex modes affected (same backend).

### LINKED ISSUE #629: mcp: codedb_read/edit reject absolute paths inside the project root as 'path traversal'
## Problem

`isPathSafe` (`src/mcp.zig:5075`) rejects **every** absolute path via a blanket `if (path[0] == '/') return false`. Because `handleRead`/`handleEdit` gate on `isPathSafe` before resolving anything, an absolute path that points at a file **inside the indexed project root** is rejected as `path traversal not allowed`.

This is invisible in scalar metrics but shows up in real traces. In the #626 session trajectory the agent called codedb twice, got an instant (10µs) `path traversal not allowed` / `PathNotAllowed` rejection both times, then abandoned codedb and fell back to seven `bash` calls to finish the task:

```
webfetch, codedb!, codedb!, read_file!, bash, bash, bash, bash, bash, bash, bash
```

Agents naturally hold and pass absolute paths. A bare rejection with no remediation makes them drop the tool — codedb's correctness is masked by bash fallback, so the task still "passes" while codedb contributed nothing.

## Failing Test

`src/test_mcp.zig` (run `zig build test-mcp`):

```zig
test "issue-629: isPathSafe rejects absolute paths inside the project root" {
    const MCP = @import("mcp.zig");

    var buf: [std.fs.max_path_bytes]u8 = undefined;
    const root_len = try std.Io.Dir.cwd().realPathFile(io, ".", &buf);
    const root = buf[0..root_len];

    const abs = try std.fs.path.join(testing.allocator, &.{ root, "src", "main.zig" });
    defer testing.allocator.free(abs);

    // An absolute path inside the project root is safe to read.
    try testing.expect(MCP.isPathSafe(abs));

    // Security must still hold: paths outside the project root stay rejected.
    try testing.expect(!MCP.isPathSafe("/etc/passwd"));
    try testing.expect(!MCP.isPathSafe("../../../etc/passwd"));
}
```

Fails on `main` at `testing.expect(MCP.isPathSafe(abs))` (123 pass, 1 fail).

## Expected

An absolute path that resolves **under an indexable project root** is accepted (and treated as the equivalent relative path). Absolute paths outside the root (`/etc/passwd`, system dirs) and `..` traversal stay rejected — the existing `issue-93` security assertions must still pass.

## Fix

Make the read/edit path-safety check root-aware instead of blanket-rejecting a leading `/`:

- If `path` is absolute, accept it only when it is an exact-or-child of an indexable root (reuse `root_policy.isExactOrChild` / `isIndexableRoot`), then strip the root prefix to the relative form the explorer expects.
- Keep rejecting `..` components, null bytes, backslashes, and out-of-root absolutes.
- Consider returning a remediation hint in the error (the allowed root, or "pass a path relative to <root>") so agents recover instead of abandoning the tool.

justrach:
Fixes two silent, agent-facing footguns surfaced from real session traces. Both have regression tests; full `zig build test` is green.

## #629 — `codedb_read`/`codedb_edit` reject absolute paths inside the project

`isPathSafe` rejected **every** absolute path via a blanket leading-`/` check, so a path pointing at a file *inside* the indexed root returned `path traversal not allowed`. Agents hold absolute paths, hit the terse error, and abandon codedb for `bash` — the #626 trajectory shows `codedb!,codedb!` then seven `bash` fallbacks.

- New `projectRelPath(path, root)`: safe relative paths pass through; an absolute path is accepted only when it is exact-or-child of the project root and is rewritten to its relative form.
- `handleRead`/`handleEdit` resolve through it using cwd's realpath as the root.
- Out-of-root absolutes, `..` traversal, NUL bytes, backslashes, and sensitive files stay rejected (the existing `issue-93` security test is unchanged).
- `root_policy.isExactOrChild` is now `pub`.

## #628 — regex alternation silently returns 0 matches

A top-level alternation like `xy|createGateway` or `.*|foo`, where one branch is too short / metachar-only to yield a trigram, prefiltered candidates down to only the trigram-bearing branches — silently dropping files that matched the other branch. `mode=regex` returned 0 and read as an authoritative "not found".

- `decomposeRegex` now detects any branch with zero trigrams and returns an unconstrained query, so the search scans every file.
- Alternations where every branch has trigrams still prefilter as before.

## Tests
- `test_mcp.zig` — `issue-629: projectRelPath accepts absolute paths inside the project root`
- `test_index.zig` — `issue-628: alternation with a no-trigram branch falls back to scan-all` (verified to fail pre-fix with `expected 0, found 1`)

Closes #629
Closes #628

### DIFF
--- a/src/index.zig
+++ b/src/index.zig
@@ -2757,14 +2757,24 @@ pub fn decomposeRegex(pattern: []const u8, allocator: std.mem.Allocator) !RegexQ
     if (top_pipes.items.len > 0) {
         // Top-level alternation: merge all branch trigrams into a single OR group.
         // A file matching ANY branch's trigrams is a valid candidate.
+        //
+        // Soundness (#628): if ANY branch yields no trigrams (too short, or built
+        // from regex metachars like `.`/`.*`), that branch can match a line that
+        // contains none of the other branches' trigrams. Constraining candidates
+        // to only the trigram-bearing branches would then silently drop real
+        // matches — exactly the `a|b` "returns 0 and looks authoritative" bug.
+        // When that happens we cannot prefilter at all, so fall back to an
+        // unconstrained query (candidatesRegex returns null -> scan everything).
         var all_tris: std.ArrayList(Trigram) = .empty;
         errdefer all_tris.deinit(allocator);
 
+        var any_branch_empty = false;
         var start: usize = 0;
         for (top_pipes.items) |pipe_pos| {
             const branch = pattern[start..pipe_pos];
             const branch_tris = try extractLiteralTrigrams(branch, allocator);
             defer allocator.free(branch_tris);
+            if (branch_tris.len == 0) any_branch_empty = true;
             for (branch_tris) |tri| {
                 try all_tris.append(allocator, tri);
             }
@@ -2774,8 +2784,22 @@ pub fn decomposeRegex(pattern: []const u8, allocator: std.mem.Allocator) !RegexQ
         const last_branch = pattern[start..];
         const last_tris = try extractLiteralTrigrams(last_branch, allocator);
         defer allocator.free(last_tris);
+        if (last_tris.len == 0) any_branch_empty = true;
         for (last_tris) |tri| {
             try all_tris.append(allocator, tri);
+        }
+
+        if (any_branch_empty) {
+            // Un-prefilterable alternation: scan everything rather than miss matches.
+            all_tris.deinit(allocator);
+            all_tris = .empty;
+            const e_and = try allocator.alloc(Trigram, 0);
+            const e_or = try allocator.alloc([]Trigram, 0);
+            return RegexQuery{
+                .and_trigrams = e_and,
+                .or_groups = e_or,
+                .allocator = allocator,
+            };
         }
 
         const empty_and = try allocator.alloc(Trigram, 0);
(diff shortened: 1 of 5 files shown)


justrach:
Already on release/0.2.5826: `git cherry` shows all #628/#629 commits are patch-equivalent on the release tip. Discarding in favor of the tip.

PR #634, opened 2026-06-21
codedb: byte-exact raw read (#632) + first-class `index` command (#633)

### LINKED ISSUE #632: codedb_read: ranged output is line-number-prefixed (not byte-exact) — add a raw read mode
## Problem

`codedb_read` with a line range returns **line-number-prefixed** output — `handleRead` calls `extractLines(content, start, end, /*line_numbers=*/true, …)` (hardcoded) and also always prepends `hash:<x>\n`. So a ranged read is **not a byte-exact copy** of the source.

Consequence on the agent read path: an agent that located code via `codedb_symbol`/`codedb_search` cannot use `codedb_read`'s ranged output as the `old_string` for an exact-match editor — the `  234 | ` prefixes won't match. So it falls back to a **native file read** for the pre-edit span, and codedb only ever serves *locate*, never *read+edit*. This is the codedb half of justrach/codegraff#66 (graff routes raw reads to its native tool for exactly this reason).

## Failing test (`src/test_mcp.zig`, fails on `main`)

```zig
test "issue-632: codedb_read raw mode returns byte-exact range without line-number prefixes" {
    // ... tmpdir with "alpha\nbeta\ngamma\n", BenchContext, Explorer ...
    // args: {"path":"small.txt","line_start":1,"line_end":2,"raw":true}
    bench_ctx.runDispatch(io, alloc, .codedb_read, &args, &out, &store, &explorer, &agents);
    try testing.expect(std.mem.indexOf(u8, out.items, "alpha\nbeta") != null); // exact bytes
    try testing.expect(std.mem.indexOf(u8, out.items, " | ") == null);        // no line-num prefix
}
```

Current behavior: ranged output is `   1 | alpha\n   2 | beta`, so `" | "` is present and the second assertion fails (`error.TestUnexpectedResult`). `raw:true` is silently ignored.

## Expected

A raw/byte-exact mode (e.g. `raw: true`) returns the exact source lines for the range with **no** line-number prefix (and the hash line suppressed or kept clearly separable), so `codedb_read` can feed an exact-string edit and serve the full read→edit loop, not just locate.

## Fix (sketch)

- Read a `raw` (or `number`) bool arg in `handleRead`; pass `!raw` as `extractLines`'s `line_numbers`.
- In raw mode, omit the `hash:` header (or move it out-of-band) so the body is verbatim.
- Keep the default (prefixed) unchanged for human/inspection use.

### LINKED ISSUE #633: codedb `index` is a phantom command: scans + persists but exits 1 with 'unknown command: index'
## Problem

`codedb <root> index` is a **phantom command**. The cold-load path keys on `cmd == "index"` (it's in `needs_word_index`), so running it **does** scan + persist the trigram/word index + project snapshot — but `index` was never registered in `isCommand` and has no dispatch branch, so after silently doing the work it falls through to:

```
✗ unknown command: index   (exit 1)
```

And `codedb index` with no explicit root is a usage error for the same reason (`isCommand("index")` is false). Harnesses mask it with `… 2>&1 | tail -1`, but it's surprising and makes the documented "index then serve/mcp" flow look broken. Surfaced live in a Cube sandbox while indexing react (7,243 files): `index: ✗ unknown command: index`.

Pre-existing (the dispatch never had an `index` branch).

## Failing test (`src/test_mcp.zig`, fails on `main`)

```zig
test "issue-633: `index` is a recognized command (not a usage/unknown error)" {
    const p = main_mod.parsePositional(&[_][]const u8{ "codedb", "index" });
    try testing.expect(!p.usage_exit);
    try testing.expectEqualStrings("index", p.cmd);
    try testing.expectEqualStrings(".", p.root);
}
```

On `main`: `codedb index` (2 args) → `isCommand("index")` false → falls through → `usage_exit = true`, `cmd = ""` → `error.TestUnexpectedResult`.

## Expected

`index` is a first-class command: `codedb index` / `codedb <root> index` builds + persists the on-disk index and exits **0** with a confirmation.

## Fix

- Add `"index"` to `isCommand`'s command list (`src/cli_args.zig`).
- Add an `index` dispatch branch in `mainImpl` that reports `index ready  <N> files` and `exit(0)` (the cold-load step already did the scan + persist).

justrach:
Two codedb fixes surfaced while benchmarking codedb as an agent tool (claude -p / graff on SWE-bench), each filed with a failing `zig test` first (repo policy) then fixed in a separate commit. Full suite green; #632 also production-verified through the sandbox gateway on react (7,243 files).

## #632 — `codedb_read` raw mode (byte-exact ranged read)
Ranged `codedb_read` emitted line-number-prefixed output (`extractLines(line_numbers=true)`) + a `hash:` header, so it was not a verbatim copy of the source — agents couldn't feed it to an exact-string editor and fell back to a native read (the codedb side of justrach/codegraff#66). Adds a `raw` arg: `line_numbers=false`, no hash header, no full-file hint. Default unchanged.
- `12c5555` failing test → `1009c64` fix. Closes #632.

## #633 — `index` is a first-class command
`codedb <root> index` triggered the scan/persist path (cold-load keys on `cmd=="index"`) but wasn't in `isCommand` and had no dispatch branch, so it printed `unknown command: index` + exit 1 *after* doing the work; `codedb index` (no root) was a usage error. Registers `index` and adds a dispatch branch that reports `index ready` + exit 0.
- `2256739` failing test → `2c130a1` fix. Closes #633.

Base: `release/0.2.5826`.

### DIFF
--- a/src/cli_args.zig
+++ b/src/cli_args.zig
@@ -204,7 +204,7 @@ pub fn isValidMcpFlag(arg: []const u8) bool {
 fn isCommand(arg: []const u8) bool {
     // cli_query_cmds is the shared query-command table (see its doc); only the
     // non-query commands are listed here.
-    const commands = cli_query_cmds ++ [_][]const u8{ "snapshot", "serve", "mcp", "update", "nuke", "cli-daemon" };
+    const commands = cli_query_cmds ++ [_][]const u8{ "snapshot", "serve", "mcp", "update", "nuke", "cli-daemon", "index" };
     for (commands) |c| {
         if (std.mem.eql(u8, arg, c)) return true;
     }
--- a/src/main.zig
+++ b/src/main.zig
@@ -467,6 +467,19 @@ fn mainImpl() !void {
         try commands.runServe(&ctx);
     } else if (std.mem.eql(u8, cmd, "mcp")) {
         try commands.runMcp(&ctx);
+    } else if (std.mem.eql(u8, cmd, "index")) {
+        // #633: `index` is a first-class command. coldLoadOrScan above already
+        // scanned + persisted the on-disk index for this cmd; confirm and exit
+        // cleanly. It used to fall through to "unknown command: index" + exit 1
+        // even though the index had been built.
+        explorer.mu.lockShared();
+        const file_count = explorer.outlines.count();
+        explorer.mu.unlockShared();
+        out.p("{s}\xe2\x9c\x93{s} {s}index ready{s}  {s}{d} files{s}\n", .{
+            s.green, s.reset, s.bold, s.reset, s.dim, file_count, s.reset,
+        });
+        out.flush();
+        std.process.exit(0);
     } else {
         out.p("{s}\xe2\x9c\x97{s} unknown command: {s}{s}{s}\n", .{
             s.red, s.reset, s.bold, cmd, s.reset,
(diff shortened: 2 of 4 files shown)


justrach:
Landed on release/0.2.5826 via fast-forward (rebased onto the post-split tip; full suite green). Commits 375989e..c72a129. GitHub won't badge this 'merged' since it didn't go through the merge button, but the patches are verifiably on the release tip.

Issue #635, opened 2026-06-21
Files 512KB–1MB silently dropped from the index (contradicts the 1MB trigram intent)


justrach:
## Problem

Source files **larger than 512 KB are silently dropped from the index entirely** — no outline, no symbol, no word/search/callers/deps — and the user is never told. Only `codedb_read` reaches them (via its disk fallback).

`watcher.parseInitialScanEntry` hard-skips at **512 KB**:
```zig
// src/watcher.zig:451
if (stat.size > 512 * 1024) return null;   // not parsed, not outlined, not indexed
```
…but a few lines down the trigram threshold was *deliberately raised to 1 MB* with this rationale:
```zig
// src/watcher.zig:457-462
// Bumped from 64KB to 1MB after the search-shootout bench (... large code files like
// ReactFiberCompleteWork.js ... were invisible to substring search ...). 1MB covers all
// reasonable code files; minified/generated bundles past 1MB are correctly skipped.
const effective_skip_trigram = entry.skip_trigram or (content.len > 1024 * 1024);
```
So files **between 512 KB and 1 MB are dropped before that logic runs** — directly contradicting the stated 1 MB intent. A 600 KB hand-written or generated source file (parsers, big modules, schemas) is invisible to the whole structural surface. The same 512 KB gate is duplicated at `watcher.zig:600`, `:924`, and `hashFile` `:1066`.

Verified live: indexing a dir with a 614 KB `big.zig` reports `1 files` (only the small control), and `search`/`word`/`symbol`/`outline` all miss `big.zig`; `status` shows no "skipped" signal.

## Failing test (`src/test_index.zig`, fails on `main`)

```zig
test "issue-635: files between 512KB and 1MB are silently dropped from the index" {
    // tmp dir: src/small.zig (tinyMarker) + src/big.zig (~600KB, bigMarkerXYZ)
    try watcher.initialScanWithWorkerCount(io, &store, &explorer, root, testing.allocator, false, 1);
    // control passes:
    try testing.expect((try explorer.searchWord("tinyMarker", a)).len >= 1);
    // FAILS on main — big.zig dropped at the 512KB gate:
    try testing.expect((try explorer.searchWord("bigMarkerXYZ", a)).len >= 1);
}
```

## Expected

A 600 KB source file is indexed (outline + symbol + word + trigram), consistent with the code's own 1 MB intent. Files past the cap should be **surfaced** (e.g. a "N files skipped (too large)" line in `status`/index output) rather than silently dropped.

[...]

PR #637, opened 2026-06-22
refactor(watcher): single read gate for the index path; name the trigram cap

justrach:
## Why

#635 (PR #636) had to bump the file-size cap in **five** separate copies of the same `stat → cap → read → null-byte check` block in `watcher.zig` — and one of those copies disagreed with the trigram intent, which *was* the #635 bug. The duplication is the root cause, not the symptom. This consolidates it so the next cap change is a one-line edit.

Stacks on `issue-635-large-file-skip` (needs the `max_indexed_file_bytes` constant). Retarget to `release/0.2.5826` once #636 merges.

## What

- **One read gate.** New `readIndexableFile(io, dir, path, alloc, size, warn_oversize)` owns the cap, the read, and the binary check. The five sites (`parseInitialScanEntry`, `readFileEntry`, `indexFileOutline`, `indexFileContent`, plus the warn path) now call it. `hashFile` keeps its streaming reader (no full-buffer alloc) and is untouched.
- **Name the trigram cap.** The bare `1024 * 1024` trigram-byte threshold was duplicated across **seven** sites — the same magic-number smell. Hoisted to `max_trigram_file_bytes`, so the index cap and trigram cap are visibly paired.
- **Vectorize the binary check.** Per-site scalar `for` loop → `std.mem.indexOfScalar` (which frees on the binary-skip path, matching prior `defer`-based behavior).
- **Fix stale comments** left by #635 (`512KB` → 2 MB cap; `64KB` → trigram cap).

## Behavior

Preserving — same cap, same warn-on-oversize (only the initial-scan path warns), same binary/oversize skips. Net +9 lines but −5 duplicated blocks and −6 magic numbers.

## Tests

Full suite green (`zig build test`); `zig fmt` clean. No behavior change, so the existing #635 + index tests are the safety net.

## Not in scope (separate issues)

- Double full-file read on same-size edits (`hashFile` then `indexFileContent`) — real perf win, but touches change-detection sentinels and wants its own characterization test.

### DIFF
--- a/src/watcher.zig
+++ b/src/watcher.zig
@@ -112,6 +112,42 @@ const WorkerParsedResults = struct {
 /// trigram (see effective_skip_trigram); past this cap the file is skipped.
 /// Was 512KB, which silently dropped 512KB-1MB source files entirely.
 const max_indexed_file_bytes = 2 * 1024 * 1024;
+
+/// #635: byte threshold for the trigram index. Files up to this size get trigram
+/// coverage; larger files (up to max_indexed_file_bytes) still get outline+word
+/// indexing but skip trigrams to bound memory on large repos. Previously a bare
+/// `1024 * 1024` duplicated across seven call sites.
+const max_trigram_file_bytes = 1024 * 1024;
+
+/// #635: the single read gate every index path shares. Enforces the size cap and
+/// the binary (null-byte) check in one place, so the threshold can't drift across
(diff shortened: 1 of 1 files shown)


justrach:
Landed on `release/0.2.5826` via fast-forward alongside #636 (commit d4247ee). Closing — base branch now contains these commits.

PR #638, opened 2026-06-22
perf(watcher): read changed files once on the live-update path

justrach:
## Why

`incrementalDiff` and `drainNotifyFile` each **read a changed file twice**: once to hash it (change detection), then again to index it. After #635 widened the cap to 2 MB, that's up to **2× max_indexed_file_bytes of IO per change**. `drainNotifyFile` (the muonry instant re-index, fired on *every* edit) double-read unconditionally.

Stacks on `refactor-watcher-read-gate` (#637) — uses its `readIndexableFile` / `max_trigram_file_bytes`. Retarget to `release/0.2.5826` once the stack merges.

## What

- **`hashAndIndexFile`** — reads a file once, hashes it, and indexes from the same buffer. Returns the hash to store (`0` = skipped/over-cap, `maxInt` = IO error).
- **`indexContentBuffer`** — the shared "index a text buffer" primitive (binary check + trigram decision), reused by `indexFileContent` and `hashAndIndexFile` so the rules stay in one place.
- **`incrementalDiff`** reads the changed file once. When there's no prior hash (first edit since startup — the common case, since the in-memory `known` map starts at `hash=0`) it skips the separate hash pass entirely, and now stores a real content hash for **size-changed** files too (previously `0`), so a later revert-to-same-size is detected.
- **`drainNotifyFile`** drops its index-then-hash double read.
- **`hashFile` deleted** — its only two callers are replaced.

## Read counts

| case | before | after |
|------|--------|-------|
| first edit since startup (common) | 2 reads | **1** |
| same-size content change | 2 reads | **1** |
| identical content (touch / `git checkout`) | 1 read | 1 |
| muonry notify (every edit) | 2 reads | **1** |

## Behavior

Change detection preserved: same identical-content short-circuit, same "binary files are hashed but not indexed" split, same skip/IO-error sentinels. `std.hash.Wyhash.hash(0, buf)` over the whole buffer equals the prior 16 KB-chunked stream, so stored hashes stay comparable. The one intentional change is storing a real hash for size-changed files (strictly more correct).

## Tests

Full suite green (`zig build test`); `zig fmt` clean. No new unit test: `incrementalDiff`/`drainNotifyFile` are private and only drivable through real mtime changes, which makes a unit test inherently flaky (sub-ms edits don't bump mtime). This is an internal, behavior-preserving change with no observable API surface — the existing index/snapshot suites are the safety net (same rationale as #637).

### DIFF
--- a/src/watcher.zig
+++ b/src/watcher.zig
@@ -1084,26 +1084,40 @@ pub fn incrementalLoop(io: std.Io, store: *Store, explorer: *Explorer, queue: *E
     }
 }
 
-fn hashFile(io: std.Io, dir: std.Io.Dir, path: []const u8, size: u64) !u64 {
-    // Returns 0 for intentional skip (large files, filtered extensions).
-    // Returns maxInt(u64) on IO error so the value always differs from a valid
-    // previously stored hash of 0, preventing a false "content unchanged" conclusion.
+
+/// Index already-read file content: skip binary (a null byte in the first 512
+/// bytes), then index with or without trigrams by size. The single place that
+/// turns a text buffer into index entries — shared by indexFileContent and
+/// hashAndIndexFile so the binary + trigram rules live in one spot.
+fn indexContentBuffer(explorer: *Explorer, path: []const u8, content: []const u8, skip_trigram: bool) !void {
+    const check_len = @min(content.len, 512);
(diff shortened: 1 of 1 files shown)


justrach:
Landed on `release/0.2.5826` via fast-forward alongside #636/#637 (commit 16e5fe9, now the release tip). Closing — base branch now contains these commits.

PR #643, opened 2026-07-03
fix(root_policy): allow /var paths when temp indexing is opted in

### LINKED ISSUE #642: `--allow-temp` flag does not work for paths under /var/ on OSTree systems (Fedora Silverblue/Nobara)
## Environment

- **codedb version**: 0.2.5825
- **OS**: Fedora Silverblue / Nobara (OSTree-based)
- **Filesystem**: `/home/` is a bind mount of `/var/home/`
- **Home directory**: `/var/home/user` (also accessible as `/home/user`)

## Problem

The `--allow-temp` flag and `CODEDB_ALLOW_TEMP` environment variable only bypass the "temporary root" check for paths under `/tmp/`, but **not** for paths under `/var/`.

On OSTree-based Fedora variants, `/var/home/` is the **real** home directory — it is NOT a temporary location. This is a standard, permanent path on these systems.

### Steps to reproduce

```shell
# /var/home/ is the actual home on OSTree. This fails:
$ codedb /var/home/user status
✗ refusing to index temporary root: /var/home/user

# Should bypass with --allow-temp, but DOES NOT:
$ codedb --allow-temp /var/home/user status
✗ refusing to index temporary root: /var/home/user

# /tmp/ bypass works correctly:
$ codedb /tmp/test status
✗ refusing to index temporary root: /tmp/test

$ codedb --allow-temp /tmp/test status
✓ codedb 0.2.5825  (works)
```

### MCP mode impact

When codedb runs in MCP mode, it uses its `cwd` as the root. If the workspace is under `/var/home/...` (the norm on OSTree systems), MCP mode rejects it immediately:

```
codedb mcp: rejected root "/var/home/user/project" (denied by policy)
```

The `CODEDB_ALLOW_TEMP=true` environment variable in the MCP server config does not help.

## Workaround

Using `/home/user/` instead of `/var/home/user/` works because codedb validates the raw argument string — it does **not** resolve `realpath()` before the temporary-root check:

```shell
# Works — codedb does not detect the bind mount equivalence:
$ codedb /home/user/project status
✓ codedb 0.2.5825  ⚡ 55.9µs
  root      /home/user/project
  files     N indexed
```

However, this workaround is fragile — it depends on the fact that codedb does not canonicalize the path before checking, and it forces users to manually substitute the path prefix.

## Expected behavior

1. `--allow-temp` should also bypass the `/var/` temporary-root check, OR
2. The temporary-root detection should canonicalize the path (via `realpath()`) before checking, so bind mounts like `/var/home/` → `/home/` are recognized as non-temporary, OR
3. The `/var/` check should have exceptions for `/var/home/` which is a legitimate home directory on OSTree systems.

## Additional context

This affects Fedora Silverblue, Nobara, and any OSTree-based Linux distribution where `/home` → `/var/home`. It is not an exotic setup — these are production desktop Linux distributions with growing adoption.

XaviCode1000:
## Summary

Move `/var` and `/private/var` from the unconditional `system_prefixes` block list into the `tempIndexingAllowed()` guard, matching the existing behavior of `/tmp` and `/private/tmp`.

This lets Fedora Silverblue and CoreOS users index projects under `/var/home/...` by setting `CODEDB_ALLOW_TEMP=1`, without compromising the safe default.

## Motivation

On Fedora Silverblue and CoreOS, the user home directory lives at `/var/home/<user>` instead of `/home/<user>`. The current blanket block on `/var` prevents codedb from indexing any project under these homes, even when the user explicitly opts in via `CODEDB_ALLOW_TEMP`.

## Changes

| File | Change |
|------|--------|
| `src/root_policy.zig` | Move `/var` and `/private/var` from `system_prefixes` to the `!tempIndexingAllowed()` guard block |

## Test Plan

- [x] Default behavior: `/var` paths are still denied (existing + new tests)
- [x] With `CODEDB_ALLOW_TEMP=1`: `/var/home/<user>/project` is indexable
- [x] `zig build test` passes (pre-existing failures unrelated to this change)
- [x] `zig build -Doptimize=ReleaseSafe` compiles successfully

## Design Rationale

Reusing `CODEDB_ALLOW_TEMP` instead of adding a new env var keeps the configuration surface small. Users on affected distros set ONE env var (or pass `--allow-temp` once) and it covers both the `/tmp` and `/var` cases.

Closes #642

### DIFF
--- a/src/root_policy.zig
+++ b/src/root_policy.zig
@@ -17,23 +17,28 @@ pub fn tempIndexingAllowed() bool {
 pub fn isIndexableRoot(path: []const u8) bool {
     if (path.len == 0) return false;
     if (std.mem.eql(u8, path, "/")) return false;
-    // /tmp and /private/tmp are refused by default (footgun guard) but allowed
-    // when temp indexing is opted in (#538) — CI/SWE-bench harnesses clone into /tmp.
-    if (!tempIndexingAllowed()) {
-        if (isExactOrChild(path, "/private/tmp")) return false;
-        if (isExactOrChild(path, "/tmp")) return false;
-    }
 
     // OSTree distros (Fedora Silverblue/CoreOS/Nobara) bind-mount /home onto
     // /var/home, so /var/home/<user>/<project> is a real project path, not a
     // system dir — and realPathFile canonicalizes /home/<user>/<project> to it
     // (same as #406/#407 fold /etc→/private/etc, /var→/private/var). Accept it
-    // before the /var system-prefix block below, mirroring the /home + /Users
-    // home-dir rule: allow project subdirs, deny the bare home. No opt-in. (#642)
+    // before the /var checks below, mirroring the /home + /Users home-dir
+    // rule: allow project subdirs, deny the bare home. No opt-in. (#642)
     if (std.mem.startsWith(u8, path, "/var/home/")) {
         const rest = path["/var/home/".len..];
         // "user" (bare home) → deny; "user/project…" → allow.
         return std.mem.indexOfScalar(u8, rest, '/') != null;
+    }
+
+    // /tmp, /private/tmp, /var, and /private/var are refused by default
+    // (footgun guard) but allowed when temp indexing is opted in (#538, #642)
+    // — CI/SWE-bench harnesses clone into /tmp, and macOS TMPDIR resolves
+    // under /private/var/folders.
+    if (!tempIndexingAllowed()) {
+        if (isExactOrChild(path, "/private/tmp")) return false;
+        if (isExactOrChild(path, "/tmp")) return false;
+        if (isExactOrChild(path, "/private/var")) return false;
+        if (isExactOrChild(path, "/var")) return false;
     }
 
     const system_prefixes = [_][]const u8{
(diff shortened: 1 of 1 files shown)


justrach:
Thanks again for the report and the PR, @XaviCode1000 — this is landing.

What I did (using the maintainer-edit access on the PR): your branch was based on `main` and targeted `release/0.2.5825`, which has already shipped, so I retargeted the PR at `release/0.2.5828` (the next release branch) and rebased your commit onto it — authorship preserved. Two integration notes:

1. **#644 landed first** on the release branch: `/var/home/<user>/<project>` (the OSTree/Silverblue home) is now indexable **with no opt-in at all** — your normal home projects just work, no env var needed.
2. **Your change rides on top for everything else under `/var`**: `--allow-temp` / `CODEDB_ALLOW_TEMP=1` now unblocks `/var` and `/private/var` exactly like `/tmp`, which also fixes macOS `TMPDIR` (`/private/var/folders/...`) and CI workspaces under `/var/lib` for harnesses that already set the flag. One assertion in your test flipped accordingly (`/var/home/xavi/project` is now expected indexable by default), and I renamed it to `issue-642` to match the issue being fixed.

The ordering matters: the `/var/home` carve-out is checked **before** the temp-guard so the no-opt-in path can't be shadowed by the default `/var` deny.

Validation on the rebased branch: `zig build test` all green (865 tests), `e2e_mcp_test.py` 20/20. Merging — ships in 0.2.5828.

PR #645, opened 2026-07-04
feat(config): per-project disable via .codedbrc + installer --no-register (#640)

### LINKED ISSUE #640: disable auto-registration of mcp
Some of my projects should not really be indexed and `codedb` automatically creating snapshots and adding tools to the global mcp scope is very annoying (increases context when I don't really need the mcp server in some projects).

justrach:
Fixes #640.

## Problem

Two annoyances reported in #640:
1. `install.sh` auto-registers codedb at **user scope** in every detected client, so its ~23 tool schemas load in *every* project — context cost even where codedb isn't wanted.
2. codedb auto-indexes whatever project it starts in and drops a `codedb.snapshot` there (plus a `.gitignore` edit).

## Fix

Two independent escape hatches:

**Per-project: `.codedbrc` with `disable = true`** — codedb ignores the project entirely:
- MCP server stays alive and protocol-correct (initialize/ping), but `tools/list` returns `{"tools":[]}` (zero schemas → zero context) and `tools/call` is refused.
- Nothing is scanned, watched, warmed, or written — no `codedb.snapshot`, no `.gitignore` edit, no project registration in the central list.
- All entry points gate on the same probe: CLI root validation, deferred-scan triggers (roots/list handshake + cwd fallback), `ProjectCache.get` (per-call `project=` switching), and `codedb_index`.
- CLI commands print `✗ codedb is disabled for this project (.codedbrc: disable = true)` and exit 1.
- Malformed/unreadable configs count as *not disabled* — a typo can't brick unrelated commands.

**Install-time: `--no-register` / `CODEDB_NO_REGISTER=1`** — installs the binary without touching any client config; prints the per-project registration command instead (`claude mcp add codedb -s project -- codedb mcp`).

## Validation

- `zig build test`: **23/23 steps, 888/888 tests** (new: `issue-640` tests in config.zig + mcp.zig).
- Live check against the built binary: disabled project → `tools/list` = `{"tools": []}`, no snapshot created, CLI exit 1 with the message above; marker removed → 23 tools served again.
- `bash -n install.sh` clean; registration block unchanged when the flag is absent.

### DIFF
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -28,6 +28,18 @@ codedb as an MCP server in every client it can find — Claude Code, Codex,
 Gemini CLI, Cursor, opencode. It prints the exact `codedb mcp` command it
 registered.
 
+To install the binary **without touching any client config**, pass
+`--no-register` (or set `CODEDB_NO_REGISTER=1`):
+
+```bash
+curl -fsSL https://codedb.codegraff.com/install.sh | bash -s -- --no-register
+```
+
+then register it only where you want it, e.g. per-project:
+`claude mcp add codedb -s project -- codedb mcp`. To keep codedb out of an
+individual project instead, drop a `.codedbrc` with `disable = true` at that
+project's root (see §4).
+
 If you prefer to wire it up by hand, the client-specific snippets below
 all work directly.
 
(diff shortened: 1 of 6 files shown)


justrach:
Reverted in 37aa83e. Maintainer call: zero-config auto-indexing is codedb's core behavior — a per-project kill switch isn't something the project wants to carry. Apologies for the churn.

PR #648, opened 2026-07-04
fix(tests): issue-77 temp root portable on Linux

justrach:
First full Linux run of the test suite (CI only ever ran the bench on Linux) found `issue-77` hardcoding `/private/tmp/...` — macOS's canonical temp root, which doesn't exist on Linux and needs root to create, so the test failed with `AccessDenied` unconditionally there.

Fix: `linux → /tmp` (same temp-root policy class), everything else keeps `/private/tmp` byte-for-byte — macOS and Windows behavior unchanged.

Validation: macOS suite 23/23 green; Linux container re-run against this branch reported separately below.

### DIFF
--- a/src/test_mcp.zig
+++ b/src/test_mcp.zig
@@ -256,7 +256,12 @@ test "issue-60: telemetry disabled path is a no-op" {
 test "issue-77: mcp index accepts temporary-directory roots that cause pathological cache growth" {
     var tmp_name_buf: [128]u8 = undefined;
     const tmp_name = try std.fmt.bufPrint(&tmp_name_buf, "codedb-issue-77-{d}", .{@as(i64, @intCast(@divTrunc(cio.nanoTimestamp(), 1000)))});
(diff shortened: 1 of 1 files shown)


justrach:
Linux confirmation as promised — clean container, `release/0.2.5828` tip (includes this PR and #647), git clone (so `getGitHead` has a real repo):

```
Build Summary: 23/23 steps succeeded; 869/873 tests passed (4 skipped)
e2e_mcp_test.py: All tests passed
```

Zero failures — the 4 skips are the Windows-gated tests. That completes the three-platform matrix for 0.2.5828: **macOS 869/873 ✓, Windows 868/872 ✓ (real VM), Linux 869/873 ✓** — all with e2e green. Along the way the Linux runs also surfaced that the two earlier failure rounds were environment (zig not on subprocess PATH; tarball checkouts lacking .git for the getGitHead test), now documented in the harness.

PR #662, opened 2026-07-09
feat(search): surface definition site inline for exact-symbol queries (#1)

justrach:
**Fewer round-trips = fewer output tokens.** Stacked on #660 (extends the search nudge #660 rewrote).

When `codedb_search`'s query is a bare identifier that resolves to an indexed symbol, the nudge used to just say *"go call codedb_symbol"* — forcing a **second round-trip** (~85 tok response + the agent's own call-emission tokens) purely to learn where the symbol is defined. Now it surfaces the definition site(s) **inline** — `path:line (kind)`, up to 3, `+N more` — reusing the O(1) `findAllSymbols` the nudge already ran.

### Before → after (measured, `claude-code` client)
- `codedb_search "renderPlainSearch"` → previously 444 tok **+ a forced `codedb_symbol` call** (85 tok). Now the first call leads with `src/explore.zig:3688 (function)` — the second call is unnecessary.

### Test
`search: exact-symbol query surfaces the definition site inline` in `test_mcp.zig`. Full `zig build test` green.

> Base is `perf/search-latency-pr-a` (#660) — retarget to `release/0.2.5828` once #660 merges. First of a series on agent-output-token efficiency.

### DIFF
--- a/src/mcp.zig
+++ b/src/mcp.zig
@@ -1681,8 +1681,36 @@ fn appendSearchSymbolNudge(alloc: std.mem.Allocator, explorer: *Explorer, query:
         alloc.free(results);
     }
     if (results.len == 0) return;
+
+    // Fewer round-trips: `query` is an indexed symbol, so instead of nudging the
+    // agent to make a *second* codedb_symbol call, surface the definition site(s)
+    // inline. Prefer real definitions over import/comment matches (same filter as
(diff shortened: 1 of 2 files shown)


justrach:
Code merged into release/0.2.5828 as part of the combined search-stack merge (commits already in the base; GitHub reports no remaining diff). Closing to keep the PR list accurate.

PR #664, opened 2026-07-09
feat(search): unify result renderer + trim match lines (#4)

justrach:
Stacked on #662 (which is stacked on #660). **Rendering-only** — result set and order unchanged.

## What
1. **Consolidate** the ~16 duplicated `path:line[: text]` render sites in `mcp.zig` (scope/regex/glob/plain/query/bundle branches) behind one `appendMatchLine` helper — prefixes (`"  "` / `"- "` / `""`) and the scope suffix preserved; `paths_only` output byte-identical.
2. **Trim the match text** to cut agent output tokens:
   - strip leading indentation (the line number already locates the hit),
   - cap at 200 bytes with a **UTF-8-safe** cut + `…`.
3. **Cover the default path too** — `renderPlainSearch` in `explore.zig` (the mcp helper can't reach across modules), which the mcp-only consolidation missed. This is the highest-traffic path (a bare `codedb_search` query), where deep source indentation was the bulk of the waste.

## Before → after (default path, `ensureTotalCapacity`)
Every result line dropped its leading indentation, e.g.
```
- src/commands.zig:130:             scratch.ensureTotalCapacity(allocator, 256 + hits.len * 80) catch {};
+ src/commands.zig:130: scratch.ensureTotalCapacity(allocator, 256 + hits.len * 80) catch {};
```
Savings scale with hit count and indentation depth; pathological long lines are capped instead of dumping 400+ chars.

## Validation
- `zig build test` — full suite green (no test asserted exact indentation/line-text).
- Trim is applied after ranking/selection, so it cannot change which results are returned or their order.

## Provenance
`mcp.zig` consolidation from a 3-attempt swarm (2 attempts produced real work; winner's diff applied and independently re-verified). The `explore.zig` default-path completion + end-to-end measurement done directly — the swarm's mcp-only change left the most common path untrimmed, which I caught by measuring.

> Base is `feat/search-fewer-roundtrips` (#662). Retarget up the stack as each lands; ultimately → `release/0.2.5828`.

### DIFF
--- a/src/explore.zig
+++ b/src/explore.zig
@@ -3685,6 +3685,19 @@ pub const Explorer = struct {
         return res;
     }
 
+    /// Trim a source line for search output: strip leading ASCII indentation
+    /// (the line number already locates it) and cap length, to cut agent output
+    /// tokens. Mirrors mcp.zig's trimMatchText (MAX 200 bytes, UTF-8-safe cut).
+    fn trimSearchLine(line: []const u8) struct { text: []const u8, truncated: bool } {
+        var t = line;
+        while (t.len > 0 and (t[0] == ' ' or t[0] == '\t')) t = t[1..];
+        const MAX: usize = 200;
+        if (t.len <= MAX) return .{ .text = t, .truncated = false };
+        var cut: usize = MAX;
+        while (cut > 0 and (t[cut] & 0xC0) == 0x80) cut -= 1;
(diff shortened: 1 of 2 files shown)


justrach:
Code merged into release/0.2.5828 as part of the combined search-stack merge (commits already in the base; GitHub reports no remaining diff). Closing to keep the PR list accurate.

PR #667, opened 2026-07-09
feat(search): surface the definition line first within the def file (#2 pass 2)

justrach:
Stacked on #665. That PR floated the defining **file** to the top; this one makes the definition **line** render first *within* that file (the render was still showing the file's first hit line — often a comment or a call above the def).

**How:** after `lineSpans` computes per-file spans, stable-move spans whose line is a definition line (outline symbol named == query) to the front. Reorders the **output** spans only — `lineSpans`' sorted input is untouched, so offset resolution is unaffected. Same result set, better in-file order.

**Verified:** pinned eval (`scripts/rank-eval.py`) — def-file rank unchanged (9/10, no regression); `codedb_search bumpSearchGen` now leads with `explore.zig:2021: fn bumpSearchGen(...)` (the def) instead of a comment. Regression test fails on baseline, passes with fix; full `zig build test` green.

Remaining holdout (separate path, tracked in `experiments/ranking/def-first-eval.md`): a query that bails to `searchContentAuto` → `searchContent` (single-word, unranked) still doesn't demote docs.

### DIFF
--- a/src/explore.zig
+++ b/src/explore.zig
@@ -3931,6 +3931,31 @@ pub const Explorer = struct {
             // OOM building the offset table → bail to the full searchContent
             // path (caller falls through), which renders the same results.
             const n_spans = self.line_offsets.lineSpans(stats.path, content, target_lines[0..target_count], &spans) orelse return false;
+            // Def-line-first: within a file that defines the query symbol, render the
+            // definition line(s) before mere mentions — stable-moves matching spans to
+            // the front (reorders the OUTPUT spans, not lineSpans' sorted input).
+            if (stats.defines) {
+                if (self.outlines.get(stats.path)) |outline| {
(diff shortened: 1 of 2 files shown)


justrach:
Superseded by the v0.2.5831 release line.\n\nThe PR head commit 7fe53b9135328ef501d79917f8cf46a76e6cfbdc is already in c71bad7, and the same definition-line-first implementation/tests are present in the release:\nhttps://github.com/justrach/codedb/commit/c71bad77b1c722465888ac0b7778362e900d68c3\nhttps://github.com/justrach/codedb/blob/c71bad77b1c722465888ac0b7778362e900d68c3/src/explore.zig#L5088-L5112\n\nClosing this PR as superseded. The head branch is not being deleted.

PR #699, opened 2026-08-20
mcp: advertise readOnlyHint so plan-mode clients stop prompting on every call

poelzi:
## Problem

`tools/list` ships no `annotations` at all, so every client that gates on `annotations.readOnlyHint` has to assume codedb's navigation tools might have side effects.

In Claude Code that is not just an extra prompt. In plan mode the permission evaluator applies an MCP-specific gate:

```js
if (e.mcpInfo && !e.isReadOnly(g) && l.behavior === "passthrough" && mode === "plan" && ...)
    l = { behavior: "ask", decisionReason: { type: "mode", mode: "plan" } };
```

and then returns that `ask` **before** the allow-rule lookup runs. Deny and ask rules are consulted earlier, allow rules later — so in plan mode a configured `mcp__codedb__*` allow rule is never reached. The gate's only escape is `isReadOnly()`, which is defined as `annotations?.readOnlyHint ?? false`.

"Don't ask again" does not stick either: the plan-mode branch replaces the decision object wholesale, dropping the `addRules` suggestions that MCP's own `checkPermissions` normally attaches, so there is no persistence path.

Probing the server directly confirms the input side:

```
$ codedb mcp   # tools/list
codedb_outline  annotations = null
codedb_symbol   annotations = null
codedb_search   annotations = null
...
```

Net effect today: every `codedb_search` / `codedb_symbol` / `codedb_read` call prompts while planning, and no configuration can silence it.

## Fix

Add `"annotations":{"readOnlyHint":true}` to the 19 tools that only read the index.

`codedb_index` rebuilds the on-disk index, and `codedb_bundle` can dispatch it, so **neither is annotated** — they should keep prompting.

The core / slim / mini profile builders and `buildAugmentedToolsList` all derive from this same static JSON, so the hint rides along on every profile with no extra code.

## Verification

- New test `mcp: read-only tools advertise readOnlyHint on every profile` asserts the hint across `full`, `core`, `slim` and `mini`, and asserts that `codedb_index` / `codedb_bundle` never claim to be read-only.
- Confirmed over a live stdio handshake on the default mini profile (the six tools an agent client actually sees):

```
tool count: 6
  codedb_outline     annotations={'readOnlyHint': True}
  codedb_symbol      annotations={'readOnlyHint': True}
  codedb_search      annotations={'readOnlyHint': True}
  codedb_callers     annotations={'readOnlyHint': True}
  codedb_context     annotations={'readOnlyHint': True}
  codedb_deps        annotations={'readOnlyHint': True}
```

- `zig build test` passes (the one unrelated `test_index` `issue-690` inotify crash on this base is addressed in a separate PR).

Every MCP client with a plan or read-only mode benefits, not just Claude Code.


justrach:
Landed on main via merge commit 737b66c.

Resolved the conflict with 0.2.5842's tool-surface changes: annotations re-applied to every read-only entry, codedb_bundle and codedb_index left unhinted per this PR's intent, and the coverage test extended to the new codedb_explain / codedb_list_dir tools. Full `zig build test` suite green on the merge commit.

Thanks @poelzi - nice catch on plan-mode prompting.

Issue #31, opened 2026-03-24
O(n²) trigram dedup in readFromDiskInner


justrach:
## Bug

`index.zig:730-734` deduplicates trigrams per file by linear scan over the existing list. For files with N distinct trigrams this is O(N²). Large minified files can have tens of thousands of unique trigrams.

## Fix
Use a temporary `AutoHashMap` for the dedup pass.

## Location
- `src/index.zig:730-734`

PR #726, opened 2026-08-28
Fix ambiguous symbol navigation and real-repo call graphs

### LINKED ISSUE #725: Name-only find/callpath pick the wrong definition; common names drop the real entrypoint
## Summary

On a mixed Zig/Python/TS repo with a fresh structural index (codedb 0.2.5846, scan ready, ~570 outlined files), **exact unique identifiers are excellent** and **architecture `codedb_context` now ranks the right docs** after a rebuild. The remaining failure is name-only resolution:

1. `codedb find` / `codedb_explain` return **the first index hit** for a common name. For `main`, `handle`, `run`, `Capsule`, `API` that is a Python `def`, a TS `type`, or a shell `API=`, not the Zig entrypoint.
2. `codedb symbol main` / `explain main` list **exactly 50 results and stop**, with **no truncation flag**. The binary `pub fn main` in `src/main.zig` is **absent**. 49 of 50 hits are Python `def main`.
3. `codedb callpath` resolves callees **by identifier only**, then takes **any 1-hop node with that name**. `cmdServe → run` lands on `src/workspace.zig:run` even though the source is `server.run(...)`. Same-file helpers collide the same way (`enqueue → persistLocked` goes to another module's `persistLocked`).
4. Callback-style calls are invisible to `extractCallees` (`ident(` only). `std.Thread.spawn(.{}, connection, .{...})` is a `callers` hit but `callpath run connection` is none, so the HTTP accept loop never reaches `handle` in the graph.

This is the follow-up to **#718** (hybrid architecture ranking). After a cold `index` + `semantic-index`, an overview `codedb_context` query now puts `ARCHITECTURE.md`, `CODEBASE.md`, `src/main.zig`, `src/server.zig`, `src/api.zig` in the top files. Hybrid ranking is no longer the story. **Find / symbol-cap / call-graph still are.**

Related: #718 (closed), #656 (call-graph invalidation), `explore.zig` `findSymbol` (`locs.items[0]`), `codegraph.buildEdges` (fan-out to every same-named id), `isUbiquitousName` (includes `run` for *display* callees only).

## Corpus shape (redacted)

- `src/*.zig` binary: `main` → `dispatch` → `cmdServe` → `server.run` → `Thread.spawn(.{}, connection, …)` → `handle` → `gitRoute` → `handleInfoRefs`
- Many experiment Python files with `def main` / `def run` / `def handle`
- Frontend TS `export type Capsule` plus a Zig `pub const Capsule = struct`
- ~13 Zig `fn persistLocked` helpers, one per persistence module
- e2e shell `API="$BASE/..."` variables

No host paths, credentials, embedding-model ids, or vector-space hashes.

## What is still correct

Unique names work. These callpaths are exact:

| from → to | hops | landing |
|---|---|---|
| `main` → `cmdServe` | 2 | `src/main.zig` `main` → `dispatch` → `cmdServe` |
| `handle` → `gitRoute` | 1 | `src/server.zig` |
| `gitRoute` → `handleInfoRefs` | 1 | `src/server.zig` → `src/gitproto.zig` |
| `gitRoute` → `handleRpc` | 1 | same |
| `connection` → `handle` | 1 | `src/server.zig` |
| `pullsRoute` → `enqueue` | 2 | API → merge-queue |
| `find gitRoute` / `parseCapsule` / `writeJsonString` / `enqueue` / `dispatch` | — | the single Zig def |

`codedb callers` also still lists the `Thread.spawn(.{}, connection, …)` site. The index *has* the facts. Rank and graph throw them away.

### 1. `find` / `explain` first-index-wins

`findSymbol` returns `locs.items[0]`. The unit test in `src/test_explore.zig` currently *requires* that ("findSymbol should return first match via index").

| query | `find` winner | `symbol` |
|---|---|---|
| `main` | experiment Python `def main` | **50 results, all Python except one other Zig `main` under `experiments/`. `src/main.zig` missing. No "truncated" flag.** |
| `handle` | experiment Python `def handle` | 26 results; Zig HTTP `handle` is present but not first |
| `run` | experiment Python `def run` | 20 results, including the real `src/server.zig:run` *and* `src/workspace.zig:run` |
| `Capsule` | frontend `type_alias` | 3 results; Zig `struct_def` is last |
| `API` | e2e shell `API=...` | 7 shell variables; no API router |
| `route` | experiment Python `def route` | 3 results; Zig `src/api.zig` `route` is last |
| `serve` | not found | CLI command is `cmdServe`, so `serve` has no symbol |

`explain main` dumps a Python `def main` body.

### 2. `callpath` picks the wrong same-named callee

```
callpath cmdServe run
  cmdServe → run
  src/main.zig:cmdServe
    → src/workspace.zig:run          // WRONG
  // actual call is server.run(...) → src/server.zig:run

callpath enqueue persistLocked
  enqueue → persistLocked
  src/mergequeue.zig:enqueue
    → src/reviews.zig:persistLocked  // WRONG; 13 persistLocked fns exist
  // actual call is mergequeue's own persistLocked

callpath cmdServe handleInfoRefs
  no call path within 12 hops        // real path is ~5 hops if run/connection resolved

callpath run connection
  no call path within 12 hops        // spawn(.{}, connection, .{}) is not ident(
```

`codegraph.extractCallees` walks the identifier immediately before `(`. So `server.run(` becomes the callee name `run`. `buildEdges` then emits an edge to **every** node named `run` (weight `1/n`). `findCallPath` BFS is unweighted hops to **any** node with the target name, so the first `run` at distance 1 wins — here `workspace.zig`, not `server.zig`.

`resolveCallees` (used for context display) *does* skip ambiguous names and `isUbiquitousName` (which includes `"run"`). Callpath / `buildEdges` do not use that filter. The two graphs disagree.

### 3. Function values / callbacks never become edges

```zig
const t = std.Thread.spawn(.{}, connection, .{ gpa, io, rcfg, stream, &slots }) catch ...;
```

`callers connection` includes this line. `extractCallees` does not, because `connection` is not followed by `(`. The accept loop therefore never connects `run → connection → handle` in the call graph, even if `run` were disambiguated.

### 4. Related footguns (lower priority)

- `glob '*.zig'` matches **one** file (repo-root `build.zig`). Comment in `matchGlob` is explicit: a single `*` cannot cross `/`. Agents type `*.zig` expecting the whole tree; `**/*.zig` is the working pattern (43 files here).
- `tree` "sym" counts locals: `main.zig` 106 tree-sym vs ~16 outline functions. Not wrong, but agents treat it as def count.
- `semantic-index` aborted with `RepositoryChangedDuringAnnBuild` when a live daemon rewrote snapshots mid-build. Retry when idle succeeded. Operational, not ranking.

## Why this is a product bug

Agents are told to `find` / `explain` a name, then `callpath` A→B. On any real mixed-language repo:

- `explain main` is a Python experiment, not the binary.
- `callpath cmdServe run` is a workspace-sync helper, not the HTTP server.
- `callpath` from the CLI command to git-advertise is **none**, even though every hop except the ambiguous `run` and the callback `connection` is indexed.

Unique-name queries hide this in unit tests. The existing `findSymbol should return first match` test **locks in** the wrong policy.

## Suggested checks (should fail on main today)

1. **Ambiguous `run`.** Three files: `cmdServe` calls `server.run()`; `server.zig` has `pub fn run`; `workspace.zig` has `pub fn run`. `callpath cmdServe run` must land on `server.zig`, not `workspace.zig`. `find run` must not be allowed to return a single non-server hit as *the* definition without listing the others.
2. **Symbol cap.** Index 50+ `def main` plus `src/main.zig pub fn main`. `symbol main` / `explain main` must include the Zig entrypoint **or** set `truncated=true` and prefer `src/main.zig` / `pub fn main`. Silent 50-cap is a recall bug.
3. **Same-file helper.** `enqueue` in `a.zig` calls `persistLocked` in `a.zig`; `b.zig` also has `persistLocked`. `callpath enqueue persistLocked` must stay in `a.zig`.
4. **Fn pointer.** `spawn(.{}, connection, .{})` (or `const f = connection;`) should produce a `run → connection` edge, or callpath docs should say callbacks are out of scope.
5. **Language/entrypoint prior for `find`.** If `symbol` returns many hits, `find` should not pretend there is one. Prefer package entrypoint / `src/` / `pub fn` over experiment `def` / e2e `VAR=` / TS re-export, or print the collision.

## What this issue is not

- Not a request to reopen #718. Overview `codedb_context` (hybrid and `--local`) ranked architecture markdown + `src/{main,server,api}.zig` after the rebuild.
- Not a claim that unique-name `find`/`callpath` is broken. `gitRoute`, `parseCapsule`, `writeJsonString` are exact.
- Not an ANN/mmap failure. Structural tools only.
- No credentials, endpoints, host paths, or model identifiers.

justrach:
## Summary

- make common-name navigation deterministic before truncation, with explicit path scoping and honest ambiguity instead of arbitrary callpaths
- resolve exact same-file, imported receiver, Zig thread-callback, and nested Zig re-export edges without repository-wide guessing
- preserve full Swift function bodies and filter caller results to real invocation syntax
- make `reindex` atomically persist root and central snapshots so the next no-daemon query loads instead of rescanning
- focus `codedb_context` on production definitions before imports/tests/generated collisions
- update the vendored OpenPuffer backend through upstream PR #31 while retaining CodeDB's malformed-sidecar hardening and full-width recall
- version the warm CLI proxy handshake so a newly installed binary replaces a stale daemon instead of silently serving old retrieval behavior

Fixes #725.

## Validation

- `zig build test --summary all` — 31/31 build steps; 1,454 passed, 4 skipped
- `python3 scripts/e2e_mcp_test.py --binary zig-out/bin/codedb --project /Users/blackfloofie/tmp/codedb-fix-name-resolution-725` — 76/76
- `python3 scripts/fd_regression_test.py --binary zig-out/bin/codedb` — 45 descriptors at both 128 and 2,048 files (bounded, corpus-size independent)
- ReleaseFast build succeeded
- Zigrepper (945 files): `main` resolves exactly to `zigrep/src/exec/search_engine.zig:run`; warm snapshot load verified
- OpenWispr (125 files): `toggleDictation` resolves directly to `beginDictation`; Swift bodies restored; `start` caller false positives reduced
- CodeDB (660 files): production `src/main.zig:main` ranks first; scoped MCP/watcher callpaths resolve correctly; `.env*` and `*.pem` return no matches

No version bump, tag, or release is included in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ codedb hot                            # recently modified files
 | `codedb_word` | O(1) inverted index word lookup |
 | `codedb_callers` | Every call site of a symbol — word index ∩ outline scope, in one round-trip |
 | `codedb_explain` | Definition body + callers in one call (CLI aliases: `explain`, `around`) |
-| `codedb_callpath` | Shortest resolved call chain A→B (CLI alias: `path`) |
+| `codedb_callpath` | Shortest resolved call chain A→B (CLI alias: `path`); duplicate names list candidates—pass `from_path` / `to_path` (or CLI `--from-path` / `--to-path`) to select a file |
 | `codedb_context` | Task-shaped composer — Pareto-frontier hybrid retrieval by default: local BM25/symbol retrieval first, then local ANN search using a remote task embedding plus a fixed public calibration string when an explicit sidecar exists, or a bounded transient semantic rerank otherwise. Pass `semantic=local` for an on-device-only call. `format=json` adds typed provenance and retrieval-privacy metadata; `document_hops=1..2` expands linked Markdown |
 | `codedb_hot` | Most recently modified files |
 | `codedb_deps` | Typed dependency graph: imports by default, or Markdown links with `edge_type=documents`; document traversal is capped at 2 hops / 64 files |
--- a/src/bootstrap.zig
+++ b/src/bootstrap.zig
@@ -30,16 +30,16 @@ const project_file = @import("project_file.zig");
 /// A walk+stat of the tree is a few ms and early-exits on the first drifted
 /// file; conservative (returns not-stale) on any IO error.
 fn snapshotIsStale(io: std.Io, abs_root: []const u8, data_dir: []const u8, allocator: std.mem.Allocator) bool {
-    // Stat the snapshot that would actually be served — same precedence as
-    // loadBestSnapshot: in-repo first, then the central data dir.
+    // Stat the newest available snapshot. The central cache is authoritative
+    // for read-only checkouts, while a newer in-repo snapshot remains a valid
+    // explicit refresh. This must match loadBestSnapshot's mtime ordering.
     var snap_stat: @TypeOf(std.Io.Dir.cwd().statFile(io, abs_root, .{}) catch unreachable) = undefined;
     var found = false;
     for ([_][]const u8{ abs_root, data_dir }) |base| {
-        if (found) break;
         const snap_path = std.fmt.allocPrint(allocator, "{s}/codedb.snapshot", .{base}) catch return false;
         defer allocator.free(snap_path);
         if (std.Io.Dir.cwd().statFile(io, snap_path, .{})) |st| {
-            snap_stat = st;
+            if (!found or st.mtime.nanoseconds > snap_stat.mtime.nanoseconds) snap_stat = st;
             found = true;
         } else |_| {}
     }
@@ -96,6 +96,11 @@ fn loadSnapshotFileIfHeadMatches(
     return snapshot_mod.loadSnapshotValidatedFromFile(io, snapshot_file, expected_root, explorer, store, allocator);
 }
 
+fn snapshotFileMtime(io: std.Io, file: std.Io.File) ?i128 {
+    const stat = file.stat(io) catch return null;
+    return stat.mtime.nanoseconds;
+}
+
 pub fn loadBestSnapshot(
     io: std.Io,
     explorer: *Explorer,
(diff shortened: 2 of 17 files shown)


justrach:
Follow-up review caught and fixed a Windows-specific stale-daemon handoff bug in 7ba719e: the named-pipe listener now honors the same retire/yield result as POSIX. Re-review found no remaining blocker. Native tests remain green (261/265, 4 skipped), and `zig build -Dtarget=x86_64-windows` cross-compiles successfully.

justrach:
Final validation on f8d0cfa:

- `zig build test --summary all`: green (262/266 in the uncached MCP shard, 4 platform skips; all 31 build steps succeeded)
- `zig build -Doptimize=ReleaseFast`: green
- `zig build -Dtarget=x86_64-windows`: green
- MCP E2E: 76/76
- macOS FD regression: 45 descriptors at both 128 and 2048 files (bounded)
- paired benchmark (5 AB/BA pairs, frozen base corpus): provenance/parity pass; `codedb_context` +1.50% / +5.78 us, under the 10% / 50 us gate
- Zigrepper: 945 files; scoped `zigrep/src/main.zig:main -> zigrep/src/exec/search_engine.zig:run` resolves in one hop
- OpenWispr: 125 files; `toggleDictation -> beginDictation` resolves; callers are filtered correctly
- CodeDB checkout: 1247 files; scoped `src/main.zig:main -> src/mcp.zig:dispatch` resolves in six hops; qualified `watcher.incrementalLoop` callbacks are retained
- sensitive path smoke: `*.pem` and `.env*` return no indexed files

The ranking path keeps the established fast approximate graph; user-visible callpaths upgrade it to strict scoped/import-aware resolution. This preserves retrieval latency while fixing #725 correctness.

Issue #748, opened 2026-09-10
High CPU: macOS watcher walks the whole tree on unrelated events and rehashes watch-cap overflow


yxlyx:
## Problem
The macOS MCP watcher can consume excessive CPU after indexing has finished, even when indexed files are unchanged. Multiple MCP servers repeat the work independently.

## Root cause
Source references below are for release 0.2.5854, commit `4eaca89`:

- **Unconditional whole-tree reconciliation:** `src/watcher.zig:2406-2414,2483-2487` runs reconciliation after each event batch or polling timeout, including empty batches. `incrementalDiffDirty` rebuilds the parent index and walks from the root (`2578-2587`); even its unchanged-directory branch opens and recursively visits clean directories (`2740-2758,2778-2787`).
- **Polling candidates are treated as actual changes:** `max_watched` defaults to 1024 (`src/config.zig:22`). Directory watches consume the budget before file watches (`src/watcher.zig:1616-1640`). `addUnwatched` inserts every overflow path into the dirty set on every cycle (`1592-1596,2484-2485`). `dirtyForcesFileRead` then bypasses the unchanged-metadata shortcut in `applyKnownFile`, forcing content reads and hashing (`2720-2722,2625-2655`).
- **Events accelerate this work:** the two-second kqueue timeout is a maximum wait, not a minimum interval or debounce (`1983-2002`). Repeated directory events therefore repeatedly trigger the traversal and overflow reads. Skipping a file during indexing does not prevent its parent-directory event from triggering this work.
- **Per-process amplification:** each MCP server starts its own watcher (`src/commands.zig:487-535`). The project singleton lock applies to `cli-daemon`, not MCP (`src/main.zig:451-458`).

## Reproduction
1. On macOS, create a synthetic repository with enough directories and text files to exceed the default watch budget.
2. Start an explicit-root MCP server, complete initialization, and wait for `codedb_status` to report `scan: ready`.
3. With no retrieval requests or edits to indexed files, repeatedly create and delete a root-level `.lock` file. This extension is excluded by `skip_extensions` (`src/watcher.zig:2904-2912`).
4. Compare CPU usage with the quiet repository; check that the indexed file count and sequence remain unchanged.

This reproduces excessive CPU with indexing ready and no indexed-file changes. Increasing the watch budget reduces overflow work but does not eliminate whole-tree traversal on events.

## Fix direction
[...]

Issue #742, opened 2026-09-04
codedb update is failing


kesavkolla:
codedb update
✗ refusing to replace codedb 0.2.5852 with older release 0.2.56

If I have already latest version it should not compare with older release version.

justrach:
@kesavkolla — confirming the fix is included in [v0.2.5855](https://github.com/justrach/codedb/releases/tag/v0.2.5855), via [758d970](https://github.com/justrach/codedb/commit/758d970).

The updater now checks GitHub's latest release first. If a fallback reports an older version, it reports that it cannot confirm the latest release and leaves the installed binary unchanged. An explicitly requested downgrade remains rejected.

I rechecked the released macOS ARM64 binary against the [updater regression harness](https://github.com/justrach/codedb/blob/v0.2.5855/scripts/test_update_resolution.py): equal version, stale fallback, unavailable sources, and explicit downgrade all passed, with the binary hash unchanged in every case. These are controlled resolver tests; they do not establish the current availability of every update endpoint. This issue remains closed as fixed in v0.2.5855.

PR #750, opened 2026-09-11
fix: restore default semantic fallback and reduce macOS watcher work

justrach:
Fresh MCP projects in v0.2.5855 silently skip hosted semantic fallback: a missing OpenPuffer metadata file escapes as `FileNotFound` instead of `AnnIndexMissing`. This occurred on all 56 default-hybrid queries across fresh full clones of Express, Requests and Chi. The loader now preserves the fallback contract for missing metadata and graphs, including recovery after a warm cache. Local indexing starts automatically; repository-wide ANN still requires an explicit `semantic-index` build and rebuild after source changes.

The installer now requires a unique valid release checksum before replacing a binary, cleans up randomized temporary downloads, and supports `CODEDB_NO_INTEGRATIONS=1` for isolated installs. Its Claude hook no longer blocks native sed/awk with instructions to call the removed `codedb_edit` tool. Both website installer copies are synchronized.

This also retains the macOS watcher follow-up to #748: bounded recursive FSEvents on local APFS/HFS with conservative kqueue/polling fallback, deletion bookkeeping and ignore-policy invalidation. Recursive and forced-fallback paths passed 244 mutation checks. Three clients used 5.25% of one core combined in a short 4,800-source-file excluded-churn fixture; this is not a general CPU guarantee. Transient embedding retry, retrieval ranking and `ann_embed_ns` observability remain separate commits for review. Hosted Jina, calibration, hosted exact fallback and the 4× OpenPuffer mmap ANN pool are preserved.

**This PR remains a draft.** On the 56 observed full-repository questions, the release finds the expected file first in 24 and in the top five in 43 without semantic setup. The fallback error fix alone leaves those ranks unchanged. The existing ranking candidate reaches 28 first / 52 top-five without setup and 50 first / 56 top-five with explicit full-repository ANN. Four no-setup misses and six ANN first-result misses remain, and individual rankings regress (including Requests first-result count 11→9 without ANN). Automatic ANN setup/refresh is not implemented. All seven evaluation repositories are already observed; another ranking iteration needs a new frozen holdout. Earlier prepared-subset accuracy and latency diagnostics do not establish out-of-the-box behavior; some latency measurements still exceed the 10% review threshold.

A new loopback fault suite adds 27 repeatable checks without external services: rejected/rate-limited/malformed responses, timeouts, bounded connection retry, failed and killed builds, source changes during embedding, sensitive-path exclusion, and three simultaneous MCP readers during generation replacement. It reproduced a missing `HttpConnectionClosing` retry; the fix uses the existing single retry and deadline. The before-fix binary fails that case and the updated binary passes. The macOS PR workflow now runs the suite and uploads its report. This is lifecycle evidence, not ranking evidence. See the [fault-injection receipt](https://github.com/justrach/codedb/blob/0730dac/evals/results/2026-09-11-semantic-faults.json).

The added publication checks deny slab creation and make the old metadata immutable in owned macOS fixtures. Both failures preserve the old generation and remove unpublished files; rebuilding succeeds after restoring permissions. The expanded loopback suite passes 27/27 on the unchanged runtime.

A controlled Graff 0.0.296 / Grok 4.6 comparison against pinned Graphify answered six observed Requests questions correctly in both arms, with verified source citations. CodeDB used 10 retrieval calls; Graphify used 21; each made six source reads and took about 45 seconds. This is one prepared-index diagnostic pair, not a general accuracy or performance claim. Rejected setup attempts exposed imported-server contamination, a modern MCP metadata handshake incompatibility and startup catalog availability problems; only the isolated successful pair is scored. Traces also expose a CodeDB multiline-definition completeness case and Graphify truncation/ambiguity requiring additional lookups. See [comparison and remaining issues](https://github.com/justrach/codedb/blob/71d69ee/docs/graff-graphify-experiment.md) and [publication evidence](https://github.com/justrach/codedb/blob/71d69ee/evals/results/2026-09-11-semantic-publication.json).

Validation:
- Final combined runtime: 33/33 unit build steps succeeded; 1,320/1,331 tests passed with 11 skips in rerun groups, other groups cached. MCP checks: 76/76. New fault suite: 27/27 locally; GitHub-hosted execution is pending.
- Fresh full-repository probes: 56 default-budget queries. Eleven live lifecycle checks pass: roots handshake from `/`, first setup, warm cache, source edit, rebuild in-session, missing/corrupt files, restart, unreachable provider and local-only search.
- Installer checks: 14/14, including seven failure-preservation cases, successful install, native edit hooks and additive Claude/Codex registration. Live isolated installation downloaded and verified the published binary. Formatting, Python compilation and whitespace checks passed.
- Earlier frozen runtime: six lazy-startup scenarios, bounded descriptors, 244 watcher mutation checks, Intel macOS event smoke and macOS Intel/Linux/Windows cross-builds passed. Linux and Windows runtime behavior was not tested on this Mac. The released ARM64 updater passed four controlled resolver scenarios for #742.

See [first-use findings and remaining gaps](https://github.com/justrach/codedb/blob/0730dac/docs/out-of-box.md), [portable full-repository evidence](https://github.com/justrach/codedb/blob/273e830/evals/results/2026-09-11-out-of-box.json), [watcher measurements](https://github.com/justrach/codedb/blob/273e830/docs/watcher-748.md), and [earlier ranking/latency evidence](https://github.com/justrach/codedb/blob/273e830/docs/adr/0012-followup-retrieval-and-transport.md). Fresh-project tests reused existing machine enrollment; fresh-account onboarding was not tested. No installed client was replaced, and this follow-up has not been released or notarized.


yousafsabir:
Two observations on the watcher commit (`0c93dc3`) from the reporter of #756, against the same 16.6k-file / 5,100-dir APFS root. Read of the diff plus measurements on main; I could not build the branch locally (zig 0.17-dev required).

**1. The periodic audit still walks the whole tree every 2 s.** `retainUnchangedChild` and the top-of-walk shortcut both return early unless `parents.affected` is set, and it is only set for `events_only` passes. The `poll_due` path calls `incrementalDiffDirty` → `events_only=false`, so every directory is still opened and every known file stat'd each cycle. The comment says this is intentional. With file watches fully armed on main (`max_watched = 20000`, no overflow reads) that walk alone measured **44% of one core** here over 45 s, so I would predict roughly that after this PR on this tree, versus the 1.75% per client on the 1,200-dir fixture. While `recursive != null` a healthy FSEvents stream already reports loss via `MustScanSubDirs` / `KernelDropped` / `UserDropped`, so the audit is redundant at 2 s; backing it off to tens of seconds when the stream is healthy would remove most of the remaining steady-state cost.

**2. Excluded paths count toward the 2,048 pending cap, and cap overflow forces a whole-tree re-read.** `receive` inserts every path under the root; `shouldSkip`/`shouldSkipFile` filtering happens only after `drain`. With `FileEvents` set, a build writing into `target/` or `node_modules/` can exceed 2,048 pending paths between drains, which sets `rescan`, which marks all known files and dirs dirty and forces `dirtyForcesFileRead` on all of them. On a Rust workspace with `cargo watch` running, that is a full-tree hash per rebuild. Suggest applying the skip filter inside `receive` before counting, and on cap overflow marking directories dirty (metadata audit) rather than forcing content reads.

